PR #86, opened 2015-06-26
Fixes issue with high cpu usage with keep-alive and connection close

oglu:
There is an issue where when a keep-alive request is received and the socket is closed, the poller does not stop reading from the fd, resulting in high cpu usage due to the spare reads. This pr fixes that.

### DIFF
--- a/bjoern/server.c
+++ b/bjoern/server.c
@@ -131,10 +131,10 @@ ev_io_on_read(struct ev_loop* mainloop, ev_io* watcher, const int events)
         DBG_REQ(request, "Client disconnected");
       else
         DBG_REQ(request, "Hit errno %d while read()ing", errno);
-      close(request->client_fd);
-      ev_io_stop(mainloop, &request->ev_watcher);
-      Request_free(request);
     }
+    close(request->client_fd);
+    ev_io_stop(mainloop, &request->ev_watcher);
+    Request_free(request);
     goto out;
   }
 


jonashaag:
Thanks for the patch!

However, I don't think it's correct: If we receive `EAGAIN` it means that there's no data available *right now* (it doesn't imply the other side has disconnected). If we receive `EWOULDBLOCK` it means that the read operation can't be done because it would block and we wanted asynchronous I/O. (These cases are usually equivalent, but not necessarily.) In both cases it means that we can't read *right now*, but there might be more data available in the future. So we shouldn't close the connection just yet.

Could you come up with a short piece of code that helps me reproduce the bug?

oglu:
A very simple example would be (run in a running interpreter so the socket remains open)

```
import requests
s = requests.Session()
s.post('http://localhost:port', data={'SomeData':'a'})
s.post('http://localhost:port', data={'SomeData':'a'})
````
Cpu will jump to 100%, mostly though, context switches because of the read (it can also be observed with an strace).

Regarding the EAGAIN and EWOULDBLOCK, at least in Linux, it should not be possible to observe them as the read event is triggered when the socket is ready to be read.

jonashaag:
I still can't reproduce this. I tried Mac OS X and Linux. Could you please provide me with a complete set of scripts that show the bug? One script that starts a bjoern instance with your specific example WSGI application and one script that causes the loop.

What system are you using? Could you try to reproduce the bug on other systems as well?

jonashaag:
What version of libev are you using?

oglu:
For a completely reproducible test-case I'll have to work on it, regarding versions though

Operating system, linux, Ubuntu Trusty 14.04/Mint 17

kernel: 3.13.0-44 (plus various others in the same family)
libev: 4.15.-3

jonashaag:
This has been fixed, see #88

PR #99, opened 2016-12-12
enable listening on IPv6

toddjcrane:
small modification to allow listening on IPv6 addresses

### DIFF
--- a/bjoern.py
+++ b/bjoern.py
@@ -16,6 +16,18 @@ def bind_and_listen(host, port=None, reuse_port=False):
         # UNIX socket: "unix:/tmp/foobar.sock"
         sock = socket.socket(socket.AF_UNIX)
         sock.bind(host[5:])
+    elif hasattr(socket, 'AF_INET6') and (len(host) == 0 or ':' in host):
+        #IPv6 socket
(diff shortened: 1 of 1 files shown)


toddjcrane:
I still need to write tests for this

jonashaag:
Hi thanks for the patch. Minor things:

- how do I listen on all IPv4 interfaces? With your code it's possible to only listen on all IPv6 interfaces if I understand it correctly
- there's repeated code there that can easily be put into a function or conditional statement

Tests aren't really necessary, just look at all the other test cases, none if them are actual tears, more like examples...

toddjcrane:
To listen on all IPv4, listen on 0.0.0.0
I thought about it and decided to set the default to listen on IPv6 as listening on all IPv6 will also listen on all IPv4. Also, there was a bit of a (not so) hidden agenda to further IPv6 enable the internet. 

I'll make the refactoring changes to eliminate repetitive code and I'll have it up here shortly

holms:
Can this be confirmed please?

jonashaag:
The patch needs some minor refactoring before it can be landed.

PR #104, opened 2017-07-30
bjoern Python3 Port (tests/hello.py working)

yanghao:
Introduced a single header file "py2py3.h" to handle python2/3 difference in
bjoern and enable full python3 support on top of python2 support.

This port heavily referenced the existing bjoern-py3 implementation:
https://github.com/isaiah/bjoern-py3

but removed the bytesio.c/h completely.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,9 @@
 SOURCE_DIR	= bjoern
 BUILD_DIR	= build
+PYTHON	?= python2
 
-PYTHON_INCLUDE	= $(shell python2-config --include)
-PYTHON_LDFLAGS	= $(shell python2-config --ldflags)
+PYTHON_INCLUDE	= $(shell ${PYTHON}-config --include)
+PYTHON_LDFLAGS	= $(shell ${PYTHON}-config --ldflags)
 
 HTTP_PARSER_DIR	= http-parser
 HTTP_PARSER_OBJ = $(HTTP_PARSER_DIR)/http_parser.o
@@ -13,7 +14,7 @@ objects		= $(HTTP_PARSER_OBJ) \
 		             $(wildcard $(SOURCE_DIR)/*.c))
 
 CPPFLAGS	+= $(PYTHON_INCLUDE) -I . -I $(SOURCE_DIR) -I $(HTTP_PARSER_DIR)
-CFLAGS		+= $(FEATURES) -std=c11 -fno-strict-aliasing -fcommon -fPIC -Wall
+CFLAGS		+= $(FEATURES) -std=c99 -fno-strict-aliasing -fcommon -fPIC -Wall
 LDFLAGS		+= $(PYTHON_LDFLAGS) -l ev -shared -fcommon
 
 ifneq ($(WANT_SENDFILE), no)
@@ -40,7 +41,7 @@ small: clean
 
 _bjoernmodule:
 	@$(CC) $(CPPFLAGS) $(CFLAGS) $(LDFLAGS) $(objects) -o $(BUILD_DIR)/_bjoern.so
-	@PYTHONPATH=$$PYTHONPATH:$(BUILD_DIR) python -c "import bjoern"
+	@PYTHONPATH=$$PYTHONPATH:$(BUILD_DIR) ${PYTHON} -c "import bjoern"
 
 again: clean all
 
@@ -81,19 +82,19 @@ wget:
 	wget -O - -q -S $(TEST_URL)
 
 valgrind:
-	valgrind --leak-check=full --show-reachable=yes python2 tests/empty.py
+	valgrind --leak-check=full --show-reachable=yes ${PYTHON} tests/empty.py
 
 callgrind:
-	valgrind --tool=callgrind python2 tests/wsgitest-round-robin.py
+	valgrind --tool=callgrind ${PYTHON} tests/wsgitest-round-robin.py
 
 memwatch:
 	watch -n 0.5 \
-	  'cat /proc/$$(pgrep -n python2)/cmdline | tr "\0" " " | head -c -1; \
+	  'cat /proc/$$(pgrep -n ${PYTHON})/cmdline | tr "\0" " " | head -c -1; \
 	   echo; echo; \
-	   tail -n +25 /proc/$$(pgrep -n python2)/smaps'
+	   tail -n +25 /proc/$$(pgrep -n ${PYTHON})/smaps'
 
 upload:
-	python2 setup.py sdist upload
+	${PYTHON} setup.py sdist upload
 
 $(HTTP_PARSER_OBJ):
 	$(MAKE) -C $(HTTP_PARSER_DIR) http_parser.o CFLAGS_DEBUG_EXTRA=-fPIC CFLAGS_FAST_EXTRA=-fPIC
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ objects		= $(HTTP_PARSER_OBJ) \
 		             $(wildcard $(SOURCE_DIR)/*.c))
 
 CPPFLAGS	+= $(PYTHON_INCLUDE) -I . -I $(SOURCE_DIR) -I $(HTTP_PARSER_DIR)
-CFLAGS		+= $(FEATURES) -std=c99 -fno-strict-aliasing -fcommon -fPIC -Wall
+CFLAGS		+= $(FEATURES) -std=c11 -fno-strict-aliasing -fcommon -fPIC -Wall
 LDFLAGS		+= $(PYTHON_LDFLAGS) -l ev -shared -fcommon
 
 ifneq ($(WANT_SENDFILE), no)
@@ -40,7 +40,7 @@ small: clean
 
 _bjoernmodule:
 	@$(CC) $(CPPFLAGS) $(CFLAGS) $(LDFLAGS) $(objects) -o $(BUILD_DIR)/_bjoern.so
-	@PYTHONPATH=$$PYTHONPATH:$(BUILD_DIR) python2 -c "import bjoern"
+	@PYTHONPATH=$$PYTHONPATH:$(BUILD_DIR) python3 -c "import bjoern"
 
 again: clean all
 
--- a/Makefile
+++ b/Makefile
@@ -40,7 +40,7 @@ small: clean
 
 _bjoernmodule:
 	@$(CC) $(CPPFLAGS) $(CFLAGS) $(LDFLAGS) $(objects) -o $(BUILD_DIR)/_bjoern.so
-	@PYTHONPATH=$$PYTHONPATH:$(BUILD_DIR) python3 -c "import bjoern"
+	@PYTHONPATH=$$PYTHONPATH:$(BUILD_DIR) python -c "import bjoern"
 
 again: clean all
 
--- a/bjoern/25compat.h
+++ b/bjoern/25compat.h
@@ -2,4 +2,7 @@
 #define NOP do{}while(0)
 #define PyFile_IncUseCount(file) NOP
 #define PyFile_DecUseCount(file) NOP
+#if PY_MAJOR_VERSION >= 3
+#else
 #define PyVarObject_HEAD_INIT(type, size) PyObject_HEAD_INIT(type) size,
+#endif
--- a/bjoern/25compat.h
+++ b/bjoern/25compat.h
@@ -1,8 +1,0 @@
-#define Py_TYPE(ob) (((PyObject*)(ob))->ob_type)
-#define NOP do{}while(0)
-#define PyFile_IncUseCount(file) NOP
-#define PyFile_DecUseCount(file) NOP
-#if PY_MAJOR_VERSION >= 3
-#else
-#define PyVarObject_HEAD_INIT(type, size) PyObject_HEAD_INIT(type) size,
-#endif
--- a/bjoern/_bjoernmodule.c
+++ b/bjoern/_bjoernmodule.c
@@ -80,8 +80,6 @@ PyMODINIT_FUNC INIT_BJOERN(void)
   }
 
   PyModule_AddObject(bjoern_module, "version", Py_BuildValue("(iii)", 1, 4, 3));
-  PyModule_AddObject(bjoern_module, "FileWrapper", (PyObject *)&FileWrapper_Type);
-  PyModule_AddObject(bjoern_module, "StartResponse", (PyObject *)&StartResponse_Type);
   return bjoern_module;
 #else
   PyObject* bjoern_module = Py_InitModule("_bjoern", Bjoern_FunctionTable);
--- a/bjoern/_bjoernmodule.c
+++ b/bjoern/_bjoernmodule.c
@@ -3,7 +3,6 @@
 #include "wsgi.h"
 #include "filewrapper.h"
 
-
 static PyObject*
 run(PyObject* self, PyObject* args)
 {
@@ -44,7 +43,25 @@ static PyMethodDef Bjoern_FunctionTable[] = {
   {NULL, NULL, 0, NULL}
 };
 
-PyMODINIT_FUNC init_bjoern(void)
+#if PY_MAJOR_VERSION >= 3
+static struct PyModuleDef module = {
+	PyModuleDef_HEAD_INIT,
+	"bjoern",   /* name of module */
+	"wsgi server in C", /* module documentation, may be NULL */
+	-1,       /* size of per-interpreter state of the module,
+		     or -1 if the module keeps state
+		     in global variables. */
+	Bjoern_FunctionTable,
+	NULL, NULL, NULL, NULL,
+};
+#endif
+
+#if PY_MAJOR_VERSION >= 3
+#define INIT_BJOERN PyInit__bjoern
+#else
+#define INIT_BJOERN init_bjoern
+#endif
+PyMODINIT_FUNC INIT_BJOERN(void)
 {
   _init_common();
   _init_filewrapper();
@@ -53,7 +70,22 @@ PyMODINIT_FUNC init_bjoern(void)
   assert(FileWrapper_Type.tp_flags & Py_TPFLAGS_READY);
   PyType_Ready(&StartResponse_Type);
   assert(StartResponse_Type.tp_flags & Py_TPFLAGS_READY);
+  Py_INCREF(&FileWrapper_Type);
+  Py_INCREF(&StartResponse_Type);
 
+#if PY_MAJOR_VERSION >= 3
+  PyObject* bjoern_module = PyModule_Create(&module);
+  if (bjoern_module == NULL) {
+	  return NULL;
+  }
+
+  PyModule_AddObject(bjoern_module, "version", Py_BuildValue("(iii)", 1, 4, 3));
+  PyModule_AddObject(bjoern_module, "FileWrapper", (PyObject *)&FileWrapper_Type);
+  PyModule_AddObject(bjoern_module, "StartResponse", (PyObject *)&StartResponse_Type);
+  return bjoern_module;
+#else
   PyObject* bjoern_module = Py_InitModule("_bjoern", Bjoern_FunctionTable);
   PyModule_AddObject(bjoern_module, "version", Py_BuildValue("(iii)", 1, 4, 3));
+#endif
+
 }
--- a/bjoern/common.h
+++ b/bjoern/common.h
@@ -22,7 +22,8 @@ void _init_common(void);
 PyObject *_REMOTE_ADDR, *_PATH_INFO, *_QUERY_STRING, *_REQUEST_METHOD, *_GET,
          *_HTTP_CONTENT_LENGTH, *_CONTENT_LENGTH, *_HTTP_CONTENT_TYPE,
          *_CONTENT_TYPE, *_SERVER_PROTOCOL, *_SERVER_NAME, *_SERVER_PORT,
-         *_HTTP_1_1, *_HTTP_1_0, *_wsgi_input, *_close, *_empty_string;
+         *_HTTP_1_1, *_HTTP_1_0, *_wsgi_input, *_close, *_empty_string,
+	 *_empty_bytes;
 
 #ifdef DEBUG
   #define DBG_REQ(request, ...) \
--- a/bjoern/common.h
+++ b/bjoern/common.h
@@ -1,8 +1,6 @@
 #ifndef __common_h__
 #define __common_h__
 
-#define DEBUG 1
-
 #include <Python.h>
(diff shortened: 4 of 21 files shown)


Makefile
```diff
@@ -13,7 +13,7 @@ objects		= $(HTTP_PARSER_OBJ) \
 		             $(wildcard $(SOURCE_DIR)/*.c))
 
 CPPFLAGS	+= $(PYTHON_INCLUDE) -I . -I $(SOURCE_DIR) -I $(HTTP_PARSER_DIR)
-CFLAGS		+= $(FEATURES) -std=c99 -fno-strict-aliasing -fcommon -fPIC -Wall
+CFLAGS		+= $(FEATURES) -std=c11 -fno-strict-aliasing -fcommon -fPIC -Wall
```
jonashaag: Is this required for compilation?

Makefile
```diff
@@ -40,7 +40,7 @@ small: clean
 
 _bjoernmodule:
 	@$(CC) $(CPPFLAGS) $(CFLAGS) $(LDFLAGS) $(objects) -o $(BUILD_DIR)/_bjoern.so
-	@PYTHONPATH=$$PYTHONPATH:$(BUILD_DIR) python2 -c "import bjoern"
+	@PYTHONPATH=$$PYTHONPATH:$(BUILD_DIR) python3 -c "import bjoern"
```
jonashaag: Let's use `python` here

bjoern/25compat.h
```diff
@@ -2,4 +2,7 @@
 #define NOP do{}while(0)
 #define PyFile_IncUseCount(file) NOP
 #define PyFile_DecUseCount(file) NOP
+#if PY_MAJOR_VERSION >= 3
```
jonashaag: I think we can drop everything below 2.7, but maybe that's better suited for another PR, not sure.

bjoern/_bjoernmodule.c
```diff
@@ -53,7 +70,22 @@ PyMODINIT_FUNC init_bjoern(void)
...
+  Py_INCREF(&StartResponse_Type);
 
+#if PY_MAJOR_VERSION >= 3
+  PyObject* bjoern_module = PyModule_Create(&module);
+  if (bjoern_module == NULL) {
+	  return NULL;
+  }
+
+  PyModule_AddObject(bjoern_module, "version", Py_BuildValue("(iii)", 1, 4, 3));
+  PyModule_AddObject(bjoern_module, "FileWrapper", (PyObject *)&FileWrapper_Type);
```
jonashaag: Do we really need to export these?

bjoern/common.h
```diff
@@ -1,6 +1,8 @@
 #ifndef __common_h__
 #define __common_h__
 
+#define DEBUG 1
```
jonashaag: Forgotten debug code? :)

bjoern/py2py3.h
```diff
@@ -0,0 +1,36 @@
...
+#define _Bytes_FromStringAndSize(data, len) PyBytes_FromStringAndSize(data, len)
+#define _Unicode_FromStringAndSize(data, len) PyUnicode_FromStringAndSize(data, len)
+#define _Bytes_GET_SIZE(obj) PyBytes_GET_SIZE(obj)
+#define _Unicode_GET_SIZE(obj) PyUnicode_GET_LENGTH(obj)
+#define _Bytes_Check(obj) PyBytes_Check(obj)
+#define _Unicode_Check(obj) PyUnicode_Check(obj)
+#define _Bytes_Resize(obj, len) _PyBytes_Resize(obj, len);
+#define _Unicode_Resize(obj, len) PyUnicode_Resize(obj, len);
+#define _FromLong(n) PyLong_FromLong(n)
+#define _File_Check(file) 1 /* To Be Fixed */
```
jonashaag: Left-over TODO?

bjoern/request.c
```diff
@@ -243,31 +252,31 @@ on_message_complete(http_parser* parser)
...
+  int size = header.len+strlen("HTTP_");
+  char dest[size];
+  int i = 0;
 
-  *dest++ = 'H';
-  *dest++ = 'T';
-  *dest++ = 'T';
-  *dest++ = 'P';
-  *dest++ = '_';
+  dest[i++] = 'H';
```
jonashaag: You may also use `char dest[size] = {'H', 'T', ...}`

Makefile
yanghao: I never get the Makefile working for me, but I did need to change to c11 in the setup.py otherwise I face an issue with the struct timespec not defined issue.

Makefile
yanghao: Yep ... using python should be better.

bjoern/25compat.h
yanghao: I agree with you. I am not interested personally any version of python prior to 2.7.

bjoern/_bjoernmodule.c
yanghao: I will remove the FileWrapper. I think we should expose the version tuple. But instead of exposing it in _bjoern, it probably should better promoted to the top level bjoern module.

bjoern/common.h
yanghao: Yes ... will remove it. :-)

bjoern/py2py3.h
yanghao: This is something I wanted to check with you how to handle the PyFile_Check() ... I am not sure I fully understood this part. In my view there is no strong typing in Python so if it quarks like a duck then we assume it is a duck?

bjoern/request.c
yanghao: Yes, will do. removing 5 lines of code :-)

Makefile
yanghao: I am using ubuntu 17.04.

yanghao:
@jonashaag new changes pushed and unclear topics waiting your feedback. :-)

bjoern/py2py3.h
jonashaag: Yep, the object must have a `fileno()` method, or rather, it must return an integer file descriptor when `PyObject_AsFileDescriptor` is called with it. Checking if it is a native CPython file object was simplest to implement, but checking for `fileno()` or checking with `PyObject_AsFileDescriptor` is even better.

bjoern/request.c
```diff
@@ -252,15 +252,15 @@ on_message_complete(http_parser* parser)
...
+  int size = header.len+strlen(http_);
   char dest[size];
   int i = 0;
 
-  dest[i++] = 'H';
-  dest[i++] = 'T';
-  dest[i++] = 'T';
-  dest[i++] = 'P';
-  dest[i++] = '_';
+  while (*(http_ + i) != '\0') {
```
jonashaag: Why not simply initialise the `dest` array directly? This change doesn't make sense; it adds complexity while not  making things easier to understand

yanghao: if you do char dest[] = {...} then you end up with dest of the size of 5 (HTTP_).
if you do char dest[size] = {...} then you got a compiler error. ;-)

The size of dest is not len("HTTP_"), on top it has the header.len.

bjoern/py2py3.h
yanghao: New changes pushed. Now using hasattr() + callable() to check for fileno exists and is callable.

bjoern/request.c
jonashaag: Ah, I see, sorry. Let's use `memcpy` for the `HTTP_` part then.

Makefile
yanghao: I have blindly removed StringIO from py2 implementation and things still works. Please help to review something must be missing ... :-)

bjoern/request.c
yanghao: Done. Changes pushed. Please review. :-)

jonashaag:
Added a fix comment. Your compat module missed the `PyFile_{Inc,Dec}UseCount`, which made compilation break -- but C11 allowed compilation anyways, which is why you needed it. Now back to C99.

It still compiles even with the removed cstringio code, but it's incorrect because now the `wsgi.input` item is missing from the environment dict if no body was sent. (Check with the `tests/headers.py` script.)

By the way, I also found a segfault with the Py3 branch:

```
curl http://localhost:8080 --data adsf
```

yanghao:
I wonder how could the py3 got compiled as PyFile_{Inc,Dec}UseCount seems not available anymore in py3... I will take a look, also reproduced the segfault.

Will also take a look at the missing wsgi.input issue ...

yanghao:
It looks like the body->buf is not pointing to the correct position of the internal PyBytes buffer, using _Bytes_AS_DATA(buf) got a reasonable address. And the allocated buf in on_body() seems has body->pos equal to 0xffffffff. Any idea why?

yanghao:
It seems Py3 type definition is a bit more complex: https://docs.python.org/3/extending/newtypes.html

yanghao:
I pushed a patch that fixed the crash issue, please take a look. I am using plain malloc() buffers for the input body buffering, I should free the buffer somewhere but haven't figured out where should I do so. Sorry this is only my 4th day looking into Bjoern source code ...

yanghao:
I think I get it wrong. Now it is just a WA for not crashing, but we need a StringIO like object and present it to the server applications. Trying to use io.BytesIO for this.

jonashaag:
I'll have a look into it this evening! Thanks for all the work so far!

yanghao:
I have pushed a new change that uses io.BytesIO() to replace the CStringIO in py3, which works so far (with post data). However there is a memory leak but I cannot find if I missed some Py_DECREF() somewhere. Will continue to look into it. Py2 also builds without complain but I only tested for py3.

yanghao:
I fixed the memory leak issue (I carelessly discarded a returned PyObject ...) and now the wsgi app should see Bytes directly. stress tested with "ab -c 8 -n 1000000 -p ~/.bashrc http://localhost:8080/" and it works fine without any sign of memory leak. Now will look at the PyFile_Dec/Inc... stuff ... I think they should be replaced directly with Py_DECREF/INCREF?

yanghao:
@jonashaag Just realized you already dropped the PyFile_Inc/DecUseCount stuff and that is exactly the approach I was going to use ... :-) seems all is clear, except maybe a bit more testing (as well as performance comparison).

yanghao:
@jonashaag Do you know if this part of the python documentation is described correctly?

https://docs.python.org/3/extending/extending.html#reference-counts 
When you pass an object reference into another function, in general, the function borrows the reference from you — if it needs to store it, it will use Py_INCREF() to become an independent owner. There are exactly two important exceptions to this rule: PyTuple_SetItem() and PyList_SetItem(). These functions take over ownership of the item passed to them — even if they fail! (Note that PyDict_SetItem() and friends don’t take over ownership — they are “normal.”)

It says PyDict_SetItem borrows reference, but for CPython source code it is clearly that PyDict_SetItem() did a Py_INCREF() on the value passed to it. Did I miss anything?

bjoern/py2py3.h
```diff
@@ -0,0 +1,41 @@
...
+#define _Bytes_FromStringAndSize(data, len) PyBytes_FromStringAndSize(data, len)
+#define _Unicode_FromStringAndSize(data, len) PyUnicode_FromStringAndSize(data, len)
+#define _Bytes_GET_SIZE(obj) PyBytes_GET_SIZE(obj)
+#define _Unicode_GET_SIZE(obj) PyUnicode_GET_LENGTH(obj)
+#define _Bytes_Check(obj) PyBytes_Check(obj)
+#define _Unicode_Check(obj) PyUnicode_Check(obj)
+#define _Bytes_Resize(obj, len) _PyBytes_Resize(obj, len)
+#define _Unicode_Resize(obj, len) PyUnicode_Resize(obj, len)
+#define _FromLong(n) PyLong_FromLong(n)
+#define _Callable_Check(obj)
```
jonashaag: This looks like dead code?

bjoern/request.c
```diff
@@ -174,24 +182,25 @@ on_headers_complete(http_parser* parser)
...
-    PyObject* buf = PyString_FromStringAndSize(NULL, parser->content_length);
-    body = (Iobject*)PycStringIO->NewInput(buf);
-    Py_XDECREF(buf);
-    if(body == NULL)
-      return 1;
-    _set_header(_wsgi_input, (PyObject*)body);
-    Py_DECREF(body);
+    body = PyObject_CallMethodObjArgs(IO, STR_bytesio, NULL);
+    if (body == NULL) {
+	    printf("call io.BytesIO failed\n");
```
jonashaag: Either make this a real `DEBUG()` call or get rid of it entirely.

setup.py
```diff
@@ -12,7 +12,7 @@
     include_dirs  = ['http-parser', '/usr/include/libev'],
     define_macros = [('WANT_SENDFILE', '1'),
                      ('WANT_SIGINT_HANDLING', '1')],
-    extra_compile_args = ['-std=c99', '-fno-strict-aliasing', '-fcommon',
+    extra_compile_args = ['-std=c11', '-fno-strict-aliasing', '-fcommon',
```
jonashaag: Did this become mandatory again?

jonashaag:
OK, I'll need a bit more time as the changes are bigger than I expected. Thanks for all your work so far! Expect a review in the next few days

bjoern/py2py3.h
yanghao: Ooops, will remove.

bjoern/request.c
yanghao: Get rid of it. as it has a none zero return value. Also changed another place printf()->PyErr_SetString().

setup.py
yanghao: The Makefile doesn't work at all from my side. And if in setup.py I do not enable c11 I got a warning:
x86_64-linux-gnu-gcc -pthread -DNDEBUG -g -fwrapv -O2 -Wall -Wstrict-prototypes -g -fdebug-prefix-map=/build/python3.5-7CCmgg/python3.5-3.5.3=. -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2 -fPIC -DWANT_SENDFILE=1 -DWANT_SIGINT_HANDLING=1 -Ihttp-parser -I/usr/include/libev -I/usr/include/python3.5m -c bjoern/portable_sendfile.c -o build/temp.linux-x86_64-3.5/bjoern/portable_sendfile.o -std=c99 -fno-strict-aliasing -fcommon -fPIC -Wall -Wextra -Wno-unused-parameter -Wno-missing-field-initializers -g
In file included from /usr/include/python3.5m/Python.h:65:0,
                 from bjoern/portable_sendfile.h:1,
                 from bjoern/portable_sendfile.c:3:
/usr/include/python3.5m/pytime.h:136:56: warning: ‘struct timespec’ declared inside parameter list will not be visible outside of this definition or declaration
 PyAPI_FUNC(int) _PyTime_AsTimespec(_PyTime_t t, struct timespec *ts);
                                                        ^~~~~~~~

yanghao: Makefile error:

hua@grass:~/git/bjoern $ make
/bin/sh: 1: --exec-prefix: not found
/bin/sh: 1: --includes: not found
/bin/sh: 1: --cflags: not found
/bin/sh: 1: --libs: not found
/bin/sh: 1: /bin/sh: 1: --extension-suffix: not found
--ldflags: not found/bin/sh: 1:
--configdir: not found
/bin/sh: 1: --help: not found
Makefile:52: recipe for target 'build/filewrapper.o' failed
make: *** [build/filewrapper.o] Error 127
hua@grass:~/git/bjoern $

bjoern/request.c
yanghao: Looks like PyImport_ImportModule already have the exception set, so will remove this printf() too in request init.

setup.py
jonashaag: Weird. Are you using a custom-compiled Python or some special OS?

yanghao: Nope. everything standard on ubuntu 17.04.

yanghao:
FYI, I did a bit performance comparison between python3/bjoern and nginx, and bjoern is outperforming nginx for the hello world simple request scenario when we have a lot of concurrent connections. For single threaded the two are having almost identical performance (in terms of requests/per second, latency).

tjb1982:
Looks like this will arrive just in time for a project I'm starting this week. This is a cool project. Thank you!

setup.py
yanghao: It looks like if the "-std=c99/c11" is removed completely then it also builds fine without any warnings. So I added a new patch that removed "-std=c99/c11" from setup.py.

yanghao:
@jonashaag any chance to review the code further?

jonashaag:
Hey, sorry for letting you wait for so long. I hope to find some time today to review it.

yanghao:
no problem，thanks.

setup.py
```diff
@@ -12,8 +12,8 @@
     include_dirs  = ['http-parser', '/usr/include/libev'],
     define_macros = [('WANT_SENDFILE', '1'),
                      ('WANT_SIGINT_HANDLING', '1')],
-    extra_compile_args = ['-std=c99', '-fno-strict-aliasing', '-fcommon',
-                          '-fPIC', '-Wall', '-Wextra', '-Wno-unused-parameter',
+    extra_compile_args = ['-fno-strict-aliasing',
+                          '-Wall', '-Wextra', '-Wno-unused-parameter',
                           '-Wno-missing-field-initializers', '-g']
```
jonashaag: Why change these parameters?

bjoern/request.c
```diff
@@ -217,20 +225,25 @@ on_message_complete(http_parser* parser)
...
-    /* We abused the `pos` member for tracking the amount of data copied from
-     * the buffer in on_body, so reset it to zero here. */
-    ((Iobject*)body)->pos = 0;
+    /* first do a seek(0) and then read() returns all data */
+    PyObject *buf = PyObject_CallMethodObjArgs(body, STR_seek, VALUE_zero,
+		    NULL);
+    Py_DECREF(buf); /* Discard the return value */
+    buf = PyObject_CallMethodObjArgs(body, STR_read, NULL);
+    Py_DECREF(body); /* Discard the BytesIO body object */
+    _set_header_free_value(_wsgi_input, buf); /* return the bytes to wsgi app */
```
jonashaag: `wsgi.input` should be a file object, not a bytestring!

bjoern/request.c
```diff
@@ -217,20 +225,25 @@ on_message_complete(http_parser* parser)
...
   } else {
     /* Request has no body */
-    _set_header_free_value(_wsgi_input, PycStringIO->NewInput(_empty_string));
+    _set_header(_wsgi_input, _empty_bytes);
```
jonashaag: same here

bjoern/request.c
```diff
@@ -291,8 +301,18 @@ parser_settings = {
 
 void _initialize_request_module()
 {
+    IO = PyImport_ImportModule("io");
+    if (IO == NULL) {
+	    /* PyImport_ImportModule should have exception set already */
+	    return;
+    }
+    STR_bytesio = _Unicode_FromString("BytesIO");
```
jonashaag: put these constant strings into the common module and use the `_(...)` macro (if possible)

bjoern/request.c
```diff
@@ -291,8 +301,18 @@ parser_settings = {
...
+    STR_write = _Unicode_FromString("write");
+    STR_read = _Unicode_FromString("read");
+    STR_seek = _Unicode_FromString("seek");
+    VALUE_zero = _FromLong(0);
```
jonashaag: This should be put into the common module as well. Or simply leave it inline in the `seek` call; it's not as expensive as a string allocation.

bjoern/request.c
```diff
@@ -1,20 +1,26 @@
...
 #include "request.h"
 #include "filewrapper.h"
 
+#include "py2py3.h"
+
+typedef struct {
+    char *buf;
+    Py_ssize_t pos;
+    int clean;
+} bytesio;
```
jonashaag: This struct seems unused.

setup.py
yanghao: Majority are already default compilation options from setup.py so I removed them. The -std=c99 is the one that breaks on my Ubuntu 17.04, by default setup.py/gcc uses gnu99 I think.

bjoern/request.c
yanghao: Will remove.

bjoern/request.c
yanghao: Do we need to seek it to 0 before we hand over the bytesIO object?

jonashaag: Yep!

bjoern/request.c
yanghao: Will move the common strings to common.

yanghao:
@jonashaag please see the latest change and let me know if anything need to be changed.

bjoern/request.c
```diff
@@ -305,6 +304,7 @@ void _initialize_request_module()
     STR_read = _Unicode_FromString("read");
     STR_seek = _Unicode_FromString("seek");
     VALUE_zero = _FromLong(0);
+    OBJECT_empty_bytesio = PyObject_CallMethodObjArgs(IO, STR_bytesio, NULL);
```
jonashaag: I'm not sure we can simply reuse the same BytesIO object every time. Better create a new instance for each request

yanghao: Forgot to change it in last push ... check again the latest one. thoroughly tested with python3 for both empty body and none-empty body.

jonashaag:
This looks very good now, thanks! I'll look into the compilation flag issues ASAP and then we can cut a beta release with Python 3 support!

yanghao:
Super. 

Regarding the Makefile error I had, looks like:
python-config --include needs to be changed to python-config --includes.

BR, Yanghao

jonashaag:
I fixed one minor segfault (yeah, CPython reference counting is hard to get right), did some minor code style changes and squashed everything into a single commit. I tested things with Python 2 and Python 3 and it seems to work just fine. Finally, bjoern gets Python 3 support. Woohoo! :-)

Thanks for your perseverance in working on this and fixing all the issues!

I'll cut a release in the next days.

yanghao:
Which lines are causing the segfault?

jonashaag:
If `body` is not `NULL` here: https://github.com/jonashaag/bjoern/pull/104/files#diff-82601f7e9615923779eb00a8a455f34dR234 then this https://github.com/jonashaag/bjoern/pull/104/files#diff-82601f7e9615923779eb00a8a455f34dR234 incorrectly decrements the refcount (`PyDict_GetItem` returns a *borrowed* reference, not a *new* reference)

jonashaag:
Release is out!

PR #107, opened 2017-09-01
Some minor problems

yanxurui:
1. fix a bug
2. compatible py3
  * bytes/str
  * print

Thank you for your awesome project!

### DIFF
--- a/bjoern/wsgi.c
+++ b/bjoern/wsgi.c
@@ -290,7 +290,7 @@ start_response(PyObject* self, PyObject* args, PyObject* kwargs)
      * throw away any previous status and headers. */
     Py_CLEAR(request->status);
     Py_CLEAR(request->headers);
-    request->state.response_length_unknown = false;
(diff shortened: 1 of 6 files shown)


bjoern/wsgi.c
```diff
@@ -290,7 +290,7 @@ start_response(PyObject* self, PyObject* args, PyObject* kwargs)
      * throw away any previous status and headers. */
     Py_CLEAR(request->status);
     Py_CLEAR(request->headers);
-    request->state.response_length_unknown = false;
+    request->state.response_length_unknown = true;
```
jonashaag: What's the reasoning behind this change? (Haven't thought about it too much)

yanxurui: The default value should be true which means unknown. Client complains because content-length or chunk-encoded doesn't exit in the header when running `tests/test_exc_info_reference.py`.

Issue #122, opened 2018-01-24
SERVER_PORT does not follow WSGI specs


Varbin:
The WSGI specification says, it must be a string (https://www.python.org/dev/peps/pep-3333/#environ-variables). This prevents running application wrapped with `wsgiref.validate.validator`.

Traceback:

```
Traceback (most recent call last):
  File "/usr/lib/python3.6/wsgiref/validate.py", line 153, in lint_app
    check_environ(environ)
  File "/usr/lib/python3.6/wsgiref/validate.py", line 328, in check_environ
    % (key, type(environ[key]), environ[key]))
  File "/usr/lib/python3.6/wsgiref/validate.py", line 128, in assert_
    raise AssertionError(*args)
AssertionError: Environmental variable SERVER_PORT is not a string: <class 'int'> (value: 8000)
```

Shall I fix it and write a pull request?

jonashaag:
Yes please!

Varbin:
Would you mind if I add a test if bjoern does follow WSGI specification?

jonashaag:
Yes please! I even started a WSGI test suite at some point (that's more extensive than the wsgiref one) but abandoned the project for some reason I can't recall right now. So um all for WSGI compliance tests :-)

Varbin:
The simplest implementation would be to just run a wsgiref-validated (obviously not complete but at least a reference) app in bjoern. This is now implemented with 14de36ff88b35968679e34b2b25d27dc8589fe07. While it shows a warning it seems working.

Varbin:
I guess this can be safely closed now.

PR #123, opened 2018-01-25
Fix of #122

Varbin:
As said, the pull request.

If `SERVER_NAME` or `SERVER_PORT` are not, an empty string is set instead.

### DIFF
--- a/bjoern/request.c
+++ b/bjoern/request.c
@@ -206,11 +206,9 @@ on_message_complete(http_parser* parser)
   if (REQUEST->server_info->host) {
     _set_header(_SERVER_NAME, REQUEST->server_info->host);
     if (REQUEST->server_info->port == Py_None) {
-      _set_header(_SERVER_PORT,
-                  PyUnicode_FromFormat(""));
+      _set_header(_SERVER_PORT, PyUnicode_FromFormat(""));
     } else {
-      _set_header(_SERVER_PORT,
-                  PyUnicode_FromFormat("%i", REQUEST->server_info->port));
+      _set_header(_SERVER_PORT, PyUnicode_FromFormat("%i", REQUEST->server_info->port));
     }
   }
   /* SERVER_NAME is required */
--- a/bjoern/request.c
+++ b/bjoern/request.c
@@ -205,7 +205,18 @@ on_message_complete(http_parser* parser)
   /* SERVER_NAME and SERVER_PORT */
   if (REQUEST->server_info->host) {
     _set_header(_SERVER_NAME, REQUEST->server_info->host);
-    _set_header(_SERVER_PORT, REQUEST->server_info->port);
+    if (REQUEST->server_info->port == Py_None) {
+      _set_header(_SERVER_PORT,
+                  PyUnicode_FromFormat(""));
+    } else {
+      _set_header(_SERVER_PORT,
+                  PyUnicode_FromFormat("%i", REQUEST->server_info->port));
+    }
+  }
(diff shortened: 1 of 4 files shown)


bjoern/request.c
```diff
@@ -205,7 +205,18 @@ on_message_complete(http_parser* parser)
   /* SERVER_NAME and SERVER_PORT */
   if (REQUEST->server_info->host) {
     _set_header(_SERVER_NAME, REQUEST->server_info->host);
-    _set_header(_SERVER_PORT, REQUEST->server_info->port);
+    if (REQUEST->server_info->port == Py_None) {
+      _set_header(_SERVER_PORT,
+                  PyUnicode_FromFormat(""));
+    } else {
```
Varbin: In theory neither SERVER_NAME nor SERVER_PORT must be empty. Should a sane default of "localhost" be used? HTTP_HOST should take precedence in WSGI apps. Port 80 may be used  with used domain sockets (with unix sockets the reconstructed url is wrong in every case!).

bjoern/request.c
```diff
@@ -205,7 +205,18 @@ on_message_complete(http_parser* parser)
...
+      _set_header(_SERVER_PORT,
+                  PyUnicode_FromFormat("%i", REQUEST->server_info->port));
+    }
+  }
+  /* SERVER_NAME is required */
+  else {
```
Varbin: In theory neither SERVER_NAME nor SERVER_PORT must be empty. Should a sane default of "localhost" be used? HTTP_HOST should take

jonashaag: No, let's leave them empty. This is a case not covered by PEP 333, and we better break applications completely rather than sneaking in wrong values that cause hard-to-debug issues.

bjoern/request.c
```diff
@@ -205,7 +205,18 @@ on_message_complete(http_parser* parser)
   /* SERVER_NAME and SERVER_PORT */
   if (REQUEST->server_info->host) {
     _set_header(_SERVER_NAME, REQUEST->server_info->host);
-    _set_header(_SERVER_PORT, REQUEST->server_info->port);
+    if (REQUEST->server_info->port == Py_None) {
+      _set_header(_SERVER_PORT,
```
jonashaag: I guess these environment items could be moved to `_initialize_request_module` as they never change? Also, please remove these line breaks.

Varbin: They could  be set in theory. But `_initialize_request_module` does not have "access" to the real values as the host name or port number can differ in different setups.

 As for the line breaks: I guess you mean I should write
```c
_set_header(_SERVER_PORT,  PyUnicode_FromFormat("%i", REQUEST->server_info->port));
```
instead of
```c
_set_header(_SERVER_PORT,
            PyUnicode_FromFormat("%i", REQUEST->server_info->port));
```
shouldn't I?

Varbin:
@jonashaag So is c894bb4 what you meant?

bjoern/request.c
jonashaag: > But _initialize_request_module does not have "access" to the real values as the host name or port number can differ in different setups.

You can simply pass `&info` to that function (see `_bjoernmodule.c`). They won't change while the server is running, or am I missing something?

> As for the line breaks

Yes that's what I meant. 90-100 chars line width is fine.

Varbin: You your are right, that would be possible.

Varbin:
> I guess these environment items could be moved to _initialize_request_module as they never change?

Just did so.

PR #133, opened 2018-05-30
Fix wrong http-parser usage

szabolcsbalogh:
From http-parser readme:
`In case you parse HTTP message in chunks (i.e. read() request line from socket, parse, read half headers, parse, etc) your data callbacks may be called more than once. Http-parser guarantees that data pointer is only valid for the lifetime of callback.`

When HTTP header is split into multiple chunks, for example in the middle in some header field name, we got multiple `header_field_mark` calls. There you use pointer arithmetic, with a pointer from previous call, which is invalid, or more exactly, it is pointing to some buffer, which have been overwritten between this two calls, so doesn't contains the data, it meant to be.

I can get segmentation fault on `python tests/listen.py` by running this 1-2 times against it:
```python
import socket

data = """POST /test HTTP/1.1
Host: example.com
Content-Length: 3800
content-type: text/plain;charset=UTF-8
accept: */*
referer: https://example.com/basjkdamsd?q=aaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabb
accept-encoding: gzip
cookie: asd=dddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbcc
Connection: Keep-Alive

aaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccddddddddaaaaaaaaaabbbbbbbbbbccccccccccdddddddd"""

def chunks(l, n):
    """Yield successive n-sized chunks from l."""
    for i in range(0, len(l), n):
        yield l[i:i + n]

conn = socket.create_connection(('0.0.0.0', 8080))

for c in chunks(data, 100):
    conn.sendall(c)
print(conn.recv(1000))
```

So i implemented copy and append to this callbacks.

### DIFF
--- a/bjoern/py2py3.h
+++ b/bjoern/py2py3.h
@@ -16,6 +16,7 @@
 #define _Bytes_Resize(obj, len) _PyBytes_Resize(obj, len)
 #define _FromLong(n) PyLong_FromLong(n)
 #define _Unicode_EncodeLatin1(u) PyUnicode_AsLatin1String(u)
+#define _Unicode_Concat(a, b) PyUnicode_Concat(a, b)
 
 #else
 
@@ -30,6 +31,18 @@
 #define _Bytes_Resize(obj, len) _PyString_Resize(obj, len)
 #define _FromLong(n) PyInt_FromLong(n)
 #define _Unicode_EncodeLatin1(u) (Py_INCREF(u),u)
+
+#ifdef __GNUC__
+static PyObject *_Unicode_Concat(PyObject *l, PyObject *r) __attribute__ ((unused));
+#endif
+static PyObject *_Unicode_Concat(PyObject *l, PyObject *r) {
+    PyObject *ret = l;
+
+    Py_INCREF(l);  /* reference to old left will be stolen */
+    PyString_Concat(&ret, r);
+
+    return ret ? ret:NULL;
+}
 #endif
 
 #endif /* _PY2PY3_H */
--- a/bjoern/py2py3.h
+++ b/bjoern/py2py3.h
@@ -32,9 +32,6 @@
 #define _FromLong(n) PyInt_FromLong(n)
 #define _Unicode_EncodeLatin1(u) (Py_INCREF(u),u)
 
-#ifdef __GNUC__
-static PyObject *_Unicode_Concat(PyObject *l, PyObject *r) __attribute__ ((unused));
-#endif
 static PyObject *_Unicode_Concat(PyObject *l, PyObject *r) {
     PyObject *ret = l;
 
--- a/bjoern/py2py3.h
+++ b/bjoern/py2py3.h
@@ -41,7 +41,7 @@ static PyObject *_Unicode_Concat(PyObject *l, PyObject *r) {
     Py_INCREF(l);  /* reference to old left will be stolen */
     PyString_Concat(&ret, r);
 
-    return ret ? ret:NULL;
+    return ret;
 }
 #endif
 
--- a/bjoern/request.c
+++ b/bjoern/request.c
@@ -93,20 +93,20 @@ void Request_parse(Request* request, const char* data, const size_t data_len)
     Py_DECREF(val); \
   } while(0)
 
-#define _set_or_append_header(k, val, len) \
-  do { \
-    PyObject *py_val = _Unicode_FromStringAndSize(val, len); \
-    PyObject *py_val_old = PyDict_GetItem(REQUEST->headers, k); \
-    \
-    if(py_val_old) { \
-      PyObject *py_val_new = _Unicode_Concat(py_val_old, py_val); \
-      PyDict_SetItem(REQUEST->headers, k, py_val_new); \
-      Py_DECREF(py_val_new); \
-    } else { \
-      PyDict_SetItem(REQUEST->headers, k, py_val); \
-    } \
-    Py_DECREF(py_val); \
-  } while(0)
+static void
+_set_or_append_header(PyObject* headers, PyObject* k, const char* val, size_t len) {
+    PyObject *py_val = _Unicode_FromStringAndSize(val, len);
+    PyObject *py_val_old = PyDict_GetItem(headers, k);
+
+    if (py_val_old) {
+      PyObject *py_val_new = _Unicode_Concat(py_val_old, py_val);
+      PyDict_SetItem(headers, k, py_val_new);
+      Py_DECREF(py_val_new);
+    } else {
+      PyDict_SetItem(headers, k, py_val);
+    }
+    Py_DECREF(py_val);
+}
 
 static int
 on_message_begin(http_parser* parser)
@@ -121,14 +121,14 @@ on_path(http_parser* parser, const char* path, size_t len)
 {
   if(!(len = unquote_url_inplace((char*)path, len)))
     return 1;
-  _set_or_append_header(_PATH_INFO, path, len);
+  _set_or_append_header(REQUEST->headers, _PATH_INFO, path, len);
   return 0;
 }
 
 static int
 on_query_string(http_parser* parser, const char* query, size_t len)
 {
-  _set_or_append_header(_QUERY_STRING, query, len);
+  _set_or_append_header(REQUEST->headers, _QUERY_STRING, query, len);
   return 0;
 }
 
@@ -176,7 +176,7 @@ on_header_value(http_parser* parser, const char* value, size_t len)
 {
   PARSER->last_was_data = true;
   if(!PARSER->invalid_header) {
-    _set_or_append_header(PARSER->field, value, len);
+    _set_or_append_header(REQUEST->headers, PARSER->field, value, len);
   }
   return 0;
 }
--- a/bjoern/request.c
+++ b/bjoern/request.c
@@ -37,7 +37,7 @@ void Request_reset(Request* request)
 {
   memset(&request->state, 0, sizeof(Request) - (size_t)&((Request*)NULL)->state);
   request->state.response_length_unknown = true;
-  request->parser.last_was_data = true;
+  request->parser.last_call_was_header_value = true;
   request->parser.invalid_header = false;
   _free_and_unset_if_set(request->parser.field);
 }
@@ -135,10 +135,10 @@ on_query_string(http_parser* parser, const char* query, size_t len)
 static int
 on_header_field(http_parser* parser, const char* field, size_t len)
 {
-  if(PARSER->last_was_data) {
+  if(PARSER->last_call_was_header_value) {
     _free_and_unset_if_set(PARSER->field);
     PARSER->field = _Unicode_FromStringAndSize("HTTP_", 5);
-    PARSER->last_was_data = false;
+    PARSER->last_call_was_header_value = false;
     PARSER->invalid_header = false;
   }
 
@@ -174,7 +174,7 @@ on_header_field(http_parser* parser, const char* field, size_t len)
 static int
 on_header_value(http_parser* parser, const char* value, size_t len)
 {
-  PARSER->last_was_data = true;
+  PARSER->last_call_was_header_value = true;
   if(!PARSER->invalid_header) {
     _set_or_append_header(REQUEST->headers, PARSER->field, value, len);
   }
--- a/bjoern/request.c
+++ b/bjoern/request.c
@@ -10,12 +10,6 @@ static PyObject* wsgi_base_dict = NULL;
 
 static PyObject *IO_module;
 
-#define _free_and_unset_if_set(v) \
-  do { \
-    Py_XDECREF(v); \
-    v = NULL; \
-  } while(0)
-
 Request* Request_new(ServerInfo* server_info, int client_fd, const char* client_addr)
 {
   Request* request = malloc(sizeof(Request));
@@ -28,18 +22,18 @@ Request* Request_new(ServerInfo* server_info, int client_fd, const char* client_
   request->client_addr = _Unicode_FromString(client_addr);
   http_parser_init((http_parser*)&request->parser, HTTP_REQUEST);
   request->parser.parser.data = request;
-  request->parser.field = NULL;
   Request_reset(request);
   return request;
 }
 
+/* should not be called without `Request_clean` when `request->parser.field` can contain an object */
 void Request_reset(Request* request)
 {
   memset(&request->state, 0, sizeof(Request) - (size_t)&((Request*)NULL)->state);
   request->state.response_length_unknown = true;
   request->parser.last_call_was_header_value = true;
   request->parser.invalid_header = false;
-  _free_and_unset_if_set(request->parser.field);
+  request->parser.field = NULL;
 }
 
 void Request_free(Request* request)
@@ -49,6 +43,7 @@ void Request_free(Request* request)
   free(request);
 }
 
+/* Request_reset should be called after this, to reset request->parser.field to NULL */
 void Request_clean(Request* request)
 {
   if(request->iterable) {
@@ -67,7 +62,7 @@ void Request_clean(Request* request)
   Py_XDECREF(request->iterator);
   Py_XDECREF(request->headers);
   Py_XDECREF(request->status);
-  _free_and_unset_if_set(request->parser.field);
+  Py_XDECREF(request->parser.field);
 }
 
 /* Parse stuff */
@@ -112,7 +107,6 @@ static int
 on_message_begin(http_parser* parser)
 {
   REQUEST->headers = PyDict_New();
-  PARSER->field = NULL;
   return 0;
 }
 
@@ -136,12 +130,14 @@ static int
 on_header_field(http_parser* parser, const char* field, size_t len)
 {
   if(PARSER->last_call_was_header_value) {
-    _free_and_unset_if_set(PARSER->field);
+    /* We are starting a new header */
+    Py_Clear(PARSER->field);
     PARSER->field = _Unicode_FromStringAndSize("HTTP_", 5);
     PARSER->last_call_was_header_value = false;
     PARSER->invalid_header = false;
   }
 
+  /* Ignore invalid header */
   if(PARSER->invalid_header) {
     return 0;
   }
@@ -162,6 +158,7 @@ on_header_field(http_parser* parser, const char* field, size_t len)
     }
   }
 
+  /* Append field name to the part we got from previous call */
   PyObject *field_old = PARSER->field;
   PyObject *field_new = _Unicode_FromStringAndSize(field_processed, len);
   PARSER->field = _Unicode_Concat(field_old, field_new);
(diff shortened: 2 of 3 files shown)


bjoern/py2py3.h
```diff
@@ -30,6 +31,18 @@
...
+#ifdef __GNUC__
+static PyObject *_Unicode_Concat(PyObject *l, PyObject *r) __attribute__ ((unused));
+#endif
+static PyObject *_Unicode_Concat(PyObject *l, PyObject *r) {
+    PyObject *ret = l;
+
+    Py_INCREF(l);  /* reference to old left will be stolen */
+    PyString_Concat(&ret, r);
+
+    return ret ? ret:NULL;
```
jonashaag: That's a noop right?

bjoern/py2py3.h
```diff
@@ -30,6 +31,18 @@
 #define _Bytes_Resize(obj, len) _PyString_Resize(obj, len)
 #define _FromLong(n) PyInt_FromLong(n)
 #define _Unicode_EncodeLatin1(u) (Py_INCREF(u),u)
+
+#ifdef __GNUC__
+static PyObject *_Unicode_Concat(PyObject *l, PyObject *r) __attribute__ ((unused));
```
jonashaag: What does this declaration do, why is it necessary?

bjoern/request.c
```diff
@@ -59,6 +67,7 @@ void Request_clean(Request* request)
   Py_XDECREF(request->iterator);
   Py_XDECREF(request->headers);
   Py_XDECREF(request->status);
+  _free_and_unset_if_set(request->parser.field);
```
jonashaag: I don't think we actually have to NULL the field here?

bjoern/request.c
```diff
@@ -31,7 +37,9 @@ void Request_reset(Request* request)
 {
   memset(&request->state, 0, sizeof(Request) - (size_t)&((Request*)NULL)->state);
   request->state.response_length_unknown = true;
-  request->parser.body = (string){NULL, 0};
+  request->parser.last_was_data = true;
+  request->parser.invalid_header = false;
+  _free_and_unset_if_set(request->parser.field);
```
jonashaag: I don't think we actually have to NULL the field here?

bjoern/request.c
```diff
@@ -23,6 +28,7 @@ Request* Request_new(ServerInfo* server_info, int client_fd, const char* client_
   request->client_addr = _Unicode_FromString(client_addr);
   http_parser_init((http_parser*)&request->parser, HTTP_REQUEST);
   request->parser.parser.data = request;
+  request->parser.field = NULL;
```
jonashaag: I don't think we actually have to NULL the field here?

It's done in `on_message_begin`

bjoern/request.c
```diff
@@ -31,7 +37,9 @@ void Request_reset(Request* request)
 {
   memset(&request->state, 0, sizeof(Request) - (size_t)&((Request*)NULL)->state);
   request->state.response_length_unknown = true;
-  request->parser.body = (string){NULL, 0};
+  request->parser.last_was_data = true;
```
jonashaag: Can we rename this to something along the lines of `last_call_was_header_value`?

bjoern/request.c
```diff
@@ -97,22 +93,26 @@ void Request_parse(Request* request, const char* data, const size_t data_len)
     Py_DECREF(val); \
   } while(0)
 
-
-#define _set_header_from_http_header() \
+#define _set_or_append_header(k, val, len) \
```
jonashaag: With this code getting more complex, I'd rather see it in a function than as a macro.

bjoern/request.c
```diff
@@ -121,50 +121,70 @@ on_path(http_parser* parser, const char* path, size_t len)
...
 
 static int
 on_header_field(http_parser* parser, const char* field, size_t len)
 {
-  if(PARSER->value.data) {
-    /* Store previous header and start a new one */
-    _set_header_from_http_header();
-  } else if(PARSER->field.data) {
-    UPDATE_LENGTH(field);
+  if(PARSER->last_was_data) {
```
jonashaag: Would you mind adding a comment here what this case means? Along the lines of "we saw a new header field but still have the old value buffered; put them into the environ dict".

bjoern/request.c
jonashaag: It's done in `on_message_begin`

bjoern/request.c
jonashaag: It's done in `on_message_begin`

bjoern/request.c
```diff
@@ -121,50 +121,70 @@ on_path(http_parser* parser, const char* path, size_t len)
...
   }
-  PARSER->field = (string){(char*)field, len};
-  PARSER->value = (string){NULL, 0};
+
+  char field_processed[len];
+  for(size_t i=0; i<len; i++) {
+    char c = field[i];
+    if(c == '_') {
+      // CVE-2015-0219
+      PARSER->invalid_header = true;
```
jonashaag: What do you think about simply aborting request parsing entirely in this case and responding with HTTP 400?

bjoern/request.c
```diff
@@ -121,50 +121,70 @@ on_path(http_parser* parser, const char* path, size_t len)
...
+    PARSER->invalid_header = false;
+  }
+
+  if(PARSER->invalid_header) {
     return 0;
   }
-  PARSER->field = (string){(char*)field, len};
-  PARSER->value = (string){NULL, 0};
+
+  char field_processed[len];
```
jonashaag: Why not keep the old code that first buffers everything and then calls `PyUnicode_FromString` a single team at the very end? That's easier to understand IMO

bjoern/request.c
```diff
@@ -121,50 +121,70 @@ on_path(http_parser* parser, const char* path, size_t len)
...
   return 0;
 }
 
 static int
 on_headers_complete(http_parser* parser)
 {
-  if(PARSER->field.data) {
-    _set_header_from_http_header();
-  }
+  _free_and_unset_if_set(PARSER->field);
```
jonashaag: This is missing a call to `_set_or_append_header`. For the last header of the request, no `on_header_field` is sent that normally triggers `_set_or_append_header`. So this always misses out the last header.

jonashaag:
Hey, thanks for bringing this up! Definitely an oversight from not reading the http-parser readme well enough.

bjoern/py2py3.h
szabolcsbalogh: It was inspired by https://github.com/encukou/py3c/blob/c3dc22f9d8652ef0874f65bb9e54e6c4ff9371e1/include/py3c/compat.h#L83
I was not sure, but I think too that it is a noop, so i drop it.

bjoern/py2py3.h
szabolcsbalogh: It is against warrnings, if function is not used, but as this is not a generic header, this can be dropped.

bjoern/request.c
szabolcsbalogh: good point, renamed

bjoern/request.c
szabolcsbalogh: i used macro because there was a macro before, and was thinking about inline functions. It was converted to a static function.

bjoern/request.c
szabolcsbalogh: This CVE is about django, where I found this: https://github.com/django/django/commit/316b8d49746933d1845d600314b002d9b64d3e3d#diff-f6d1c75ec606389da5af6558bf57f171R142

So they are only dropping this headers, but not the whole request.
Also the previous implementation was (https://github.com/jonashaag/bjoern/blob/master/bjoern/request.c#L250) that `wsgi_http_header` returns NULL, which was called by a macro on L103, and checked on the next line, and was ignored. So if I understand right, I didn't change the behaviour there, and as this is implemented in this way in other projects, i think it is a good idea to follow this. What do you think?

bjoern/py2py3.h
szabolcsbalogh: `python setup.py build` started to report
```gcc
...
gcc -pthread -fno-strict-aliasing -O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fexceptions -fstack-protector-strong -grecord-gcc-switches -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 -m64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -D_GNU_SOURCE -fPIC -fwrapv -DNDEBUG -O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fexceptions -fstack-protector-strong -grecord-gcc-switches -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 -m64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -D_GNU_SOURCE -fPIC -fwrapv -fPIC -DWANT_SENDFILE=1 -DWANT_SIGINT_HANDLING=1 -Ihttp-parser -I/usr/include/libev -I/usr/include/python2.7 -c bjoern/common.c -o build/temp.linux-x86_64-2.7/bjoern/common.o -std=c99 -fno-strict-aliasing -fcommon -fPIC -Wall -Wextra -Wno-unused-parameter -Wno-missing-field-initializers -g
In file included from bjoern/common.c:2:
bjoern/py2py3.h:35:18: warning: ‘_Unicode_Concat’ defined but not used [-Wunused-function]
 static PyObject *_Unicode_Concat(PyObject *l, PyObject *r) {
                  ^~~~~~~~~~~~~~~
...
```

should i return this definition?

bjoern/request.c
szabolcsbalogh: some comments added around this and others

bjoern/request.c
szabolcsbalogh: I'm always appending to PyUnicode, which I am concating. I use this in all callbacks, so i didn't want to make another logic for `char *` buffer with reallocation and so on. And as this I needed to move wsgi converter logic here, which is now processing the input buffer-like, as callbacks works too.
So it can be converted to store string (char *, size_t) instead of PyObject, and then put back the wsgi function as it was, at the moment i see it as a preference on, what should be easier to understand, the convertor, or another new memory management over buffered string.

I can do both, which do you prefer?

bjoern/request.c
szabolcsbalogh: on first `on_header_value` call, the header was set, and on next calls it was appended, so you miss the last header, if it has no data, and http-parser didn't call `on_header_value` with zero length. But by https://tools.ietf.org/html/rfc7230#section-3.2 empty field data is not valid so ignoring this header should not do anything wrong.

bjoern/request.c
jonashaag: Agreed.

bjoern/request.c
jonashaag: Let's keep it this way. Since we're using unicode concatenation for header values as well, which makes sense from a simplicity and performance point of view.

bjoern/request.c
jonashaag: You're right! It was necessary in the old code because field and value were buffered before being put into the dict, but your code puts the header value into the dict on the first `on_header_value` call. So that's fine!

jonashaag:
This looks good for me now, thanks!

szabolcsbalogh:
Could I ask you to make a new release? thanks

jonashaag:
@szabolcsbalogh done!

szabolcsbalogh:
thanks a lot

PR #154, opened 2019-02-28
Add `/opt/local/include` to include_dirs

rbanffy:
This fixes compilation issues under MacPorts, since that's where libev's headers are placed.

Tested on macOS Mojave and Python 3.7 installed from MacPorts.


rbanffy:
@jonashaag, can you check this?

Issue #146, opened 2018-11-08
UNIX communication sockets left in filesystem after receiving a signal


rkrell:
I'm running Bjoern embedded in Python in this way:
```python
logger.info('Bjoern bound to UNIX socket %s', sock_name)
try:
    os.unlink(sock_name)
except OSError:
    logger.error('Could not remove previous UNIX socket %s', sock_name)
    pass
bjoern.run(api_app, 'unix:{}'.format(sock_name))
try:
    os.unlink(sock_name)
except OSError:
    pass
```

By default, supervisord sends SIGTERM to processes to quit them, as stated in issue 181.
Sending this signal, if I wouldn't remove the pending sockets, they prevent bjoern from starting next time, because the old socket is assumed to be in use.
I saw there is code for unlinking sockets in bjoern, but it is appearantly bypassed when receiving SIGTERM. Is there any way to make shutdown of bjoern more safe?

luketimothyjones:
I am having trouble replicating the issue, but I have implemented an atexit handler for tearing down the server socket in #155. Could someone verify that this fixes the issue?

m190:
I think that your fix will not be working. I've tried to make a simple wrapper that just unlinks the socket with the help of atexit, and it wasn't called. I think that the reason is mentioned [there](https://docs.python.org/3/library/atexit.html):

> **Note:** The functions registered via this module are not called when the program is killed by a signal not handled by Python, when a Python fatal internal error is detected, or when [`os._exit()`](https://docs.python.org/3/library/os.html#os._exit "os._exit") is called.

So, the signal seems to be a better option. Bud even with the signals, it seems that there should be a manual unlink before the run.

Issue #91, opened 2016-05-31
Gracefully exiting a Python process running Bjoern


saley89:
Hi,

I have a simple Python script wrapping a Bjoern web server around a WSGI app. In essence, it looks like this:

```
def signal_term_handler(signal, frame):
    logging.info('got SIGTERM')
    sys.exit(0)

def main():
    signal.signal(signal.SIGTERM, signal_term_handler)
    logging.info(" Starting Bjoern with WSGI app at http://{}:{}/".format(HOST, PORT))
    bjoern.run(router, HOST, PORT)
```

The server works just fine. The problem is that when sending a SIGTERM signal (e.g. using `kill -15`), the signal handler doesn't get called and the process doesn't die. When sending SIGKILL, the process does die, without calling the signal handler (this is expected). 

How can I intercept SIGTERM?

jonashaag:
Hm... the way Python implements signals makes it impossible to use them in this situation (C event-loop based server), because they are put into in an internal signal queue and the handlers are only called the next time the Python interpreter is executed, which is when a new request is being made to the server or an existing request is being processed.

I'm not sure how other event-loop based servers solve this problem, maybe by routinely invoking the Python interpreter.

Maybe you can do aside with signals altogether?

jonashaag:
This seems to be possible to work around with periodically calling into https://docs.python.org/2/c-api/exceptions.html#c.PyErr_CheckSignals

saley89:
Thanks for the quick reply and confirming our suspicions that it may be related to the C side of the library.  We will look into an alternative solution.  

It would be great if you could build in the CheckSignals code you mentioned above if that would allow us to handle signals correctly.  I can leave this open if you would like to look into it, otherwise I am happy to close the issue.

vin18:
Shouldn't SIGTERM be handled the same way SIGINT is ? For graceful shutdown ?

jonashaag:
That's correct, if it doesn't work like that then that's a separate issue.

vin18:
@jonashaag SIGTERM is not handled like SIGTERM as 
https://github.com/jonashaag/bjoern/blob/master/bjoern/server.c#L76 only handles SIGINT.

jonashaag:
I don't know, maybe the other way round (first check signals, then set interrupt),  but in either case we'd have to make sure we hold the GIL

Auha:
Are there any work arounds? I have an app that should not have any downtime.

ninjatrench:
[...]

PR #155, opened 2019-03-16
Move server socket teardown to an atexit handler

luketimothyjones:
Fix for #146

### DIFF
--- a/bjoern.py
+++ b/bjoern.py
@@ -1,3 +1,4 @@
+import atexit
 import os
 import socket
 import _bjoern
@@ -38,6 +39,18 @@ def server_run(sock, wsgi_app):
     _bjoern.server_run(sock, wsgi_app)
 
 
+def _close_server_socket():
+    # Handler for cleaning up the server socket when the process terminates
+    global _default_instance
+    
+    if sock.family == socket.AF_UNIX:
+        filename = sock.getsockname()
+        if filename[0] != '\0':
+            os.unlink(sock.getsockname())
+    sock.close()
+    _default_instance = None
+
+
 # Backwards compatibility API
 def listen(wsgi_app, host, port=None, reuse_port=False):
     """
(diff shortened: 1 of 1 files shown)


bjoern.py
```diff
@@ -38,6 +39,18 @@ def server_run(sock, wsgi_app):
     _bjoern.server_run(sock, wsgi_app)
 
 
+def _close_server_socket():
+    # Handler for cleaning up the server socket when the process terminates
+    global _default_instance
+    
+    if sock.family == socket.AF_UNIX:
```
jonashaag: Sorry for not reviewing this patch earlier, must have missed it!

Where does the `sock` variable come from? Did you even test this code once? If I'm not terribly mistaken the code doesn't work...

luketimothyjones: Yes I did, as I wound up using it in a production application  and it worked, but it appears that I missed a line when I copied it over from our codebase. My apologies for that.

jonashaag:
I see! Thanks for the fix!

I wonder what happens if you invoke `.run()` multiple times in the same Python interpreter. Not sure how atexit behaves. Also the teardown is done only when the interpreter exits, shouldn't we close the socket as soon as possible? Or better even, do both, i.e. have a teardown in a `finally` block as before this patch and an `atexit` handler?

We'll also have to check if the issue reported in #146 is even reproducible anymore with the fix in #91 – and then we'll also have to re-check when we make any changes resulting from the discussion in #91

luketimothyjones:
I don't see any commited code referenced in #91, only discussion; am I missing something?

luketimothyjones:
As for calling `run()` multiple times, [that's handled before the atexit register happens](https://github.com/jonashaag/bjoern/blob/5f38de0d950aac91db7b39efbb01bb116eef83a7/bjoern.py#L86) (it would throw a RuntimeError).

luketimothyjones:
I just reviewed the `atexit` docs and realized there's a problem with my fix.

```
The functions registered via this module are not called when the program is killed by a signal not handled by Python, when a Python fatal internal error is detected, or when os._exit() is called.
```

Replacing the `try/except` block that my original commit removed with a call to `_close_server_socket()` in the `except` clause would fix this oversight.

Additionally, I'm not super well read into Python's reference counting implementation, but if it doesn't play well with `atexit` it may be wise to pass `sock` as an argument to `_close_server_socket()` instead of relying on the `_default_instance` global existing. I'm 99% sure that the implementation as it stands will work fine, as I trust that Python handles this, but that 1% of paranoia is still there.

jonashaag:
Thanks for looking that stuff up! Let's wait for the resolution of #91 before we continue here.

luketimothyjones:
Welp, I had my head in the code and missed your most recent comment, so the fixes have been implemented regardless haha.

PR #156, opened 2019-04-17
Provide entry point for PasteDeploy interoperability

dataflake:
This small change makes it much easier to run bjoern in an environment that uses Paste-style .ini files to define settings and pipelines.

### DIFF
--- a/bjoern.py
+++ b/bjoern.py
@@ -82,8 +82,3 @@ def run(*args, **kwargs):
                 os.unlink(sock.getsockname())
         sock.close()
         _default_instance = None
-
-def serve_paste(app, global_conf, **kw):
-    kw['port'] = int(kw.get('port', '0')) or False
-    run(app, **kw)
-    return 0
--- a/bjoern.py
+++ b/bjoern.py
@@ -82,3 +82,8 @@ def run(*args, **kwargs):
                 os.unlink(sock.getsockname())
         sock.close()
         _default_instance = None
+
+def serve_paste(app, global_conf, **kw):
+    kw['port'] = int(kw.get('port', '0')) or False
+    run(app, **kw)
+    return 0
--- a/contrib.py
+++ b/contrib.py
@@ -1,0 +1,28 @@
+from bjoern import run
+
+
+def serve_paste(app, global_conf, **kw):
+    """ A handler for PasteDeploy-compatible runners.
+
+    Sample .ini configuration:
+    
+     [server:main]
+     use = egg:bjoern#main
+     host = 127.0.0.1
+     port = 8080
+     reuse_port = True
+    
+    If no host is specified bjoern will bind to all IPv4 IPs (0.0.0.0)
+    If no port is specified, bjoern will use a random port
+    reuse_port defaults to False
+    """
+    # Convert the values from the .ini file to something bjoern can work with
+    host = kw.get('host', '')
+    port = int(kw.get('port', '0')) or False
+    if kw.get('reuse_port', '').lower() in ('1', 'true'):
+        reuse_port = True
+    else:
+        reuse_port = False
+
+    run(app, host, port=port, reuse_port=reuse_port)
+    return 0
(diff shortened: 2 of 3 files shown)


jonashaag:
Thanks for the patch! Could this live in a bjoern.contrib module?

dataflake:
What's a bjoern.contrib module?

dataflake:
I guess you mean either a different module inside bjoern or a different egg altogether? Yes, it could live anywhere. If it is anywhere within the bjoern egg then the entry point definition in setup.py has to stay. It could even move into a separate egg entirely, but  that seems like a whole lot of scaffolding for a tiny function wrapper.

jonashaag:
Just a different file in the repository, not a different egg altogether. Entry point is fine.

dataflake:
Sure, I'll change that some time today.

dataflake:
I moved the code out into a separate module. Please take a look when you have a minute.

Thanks!

jonashaag:
Does this install as `bjoern.contrib`? Does this install at all with the current `setup.py`?

dataflake:
It doesn't install anything, it just declares an entry point. I'm not sure I understand your question. The change to the current setup.py is part of this PR, and yes, I named the new file `contrib.py` to keep it generic for other contributions.

dataflake:
P.S.: This is tested manually using `zc.buildout` where the `bjoern` egg is based on the checkout.

jonashaag:
Maybe I'm not familiar with the Paste ecosystem. Does a Paste deployment rely on having the contrib module installed when you do `pip install bjoern`? If so you'd have to add it to `setup.py` somehow, right?

dataflake:
The `contrib` module doesn't need installing. It just sits in the `bjoern` egg. `PasteDeploy` generates the correct import paths during its own processing.

dataflake:
I can see what you mean now, you can't manually import the module. But for the purposes of the entry point it does work.

jonashaag:
Am I correct that the contrib module is not installed when you do pip/setup.py install? So Paste deployment works from bjoern source only?

dataflake:
That's correct.

By the way, the entry_points things has nothing to do with PasteDeploy per se. It's a setuptools function. How that is handled behind the scenes I don't know. PasteDeploy just utilizes it.

jonashaag:
Do Paste people not use pip? Doesn’t it make sense to have pip install the contrib module as well?

dataflake:
I'm curious why you go on about "installing" something. The proposed solution is a very small enhancement and doesn't require any installation.

If you are pushing back because for some reason you don't want the code in the `bjoern` module at all then please let me know.

dataflake:
P.S.: I'm not representing Paste here. I'm representing Zope, which can easily utilize WSGI servers that have PasteDeploy-compatible entry points in their setup configuration. Composing WSGI pipelines in PasteDeploy-style .ini files seems a widely-used standard. I'm just trying to integrate support for it in `bjoern` itself. Like I said, if you don't want it just let me know and I'll close the PR and do it all externally.

jonashaag:
I am wondering how users that install bjoern using pip will get access to the Paste entrypoint, in the first place. Currently the contrib module doesn’t ship with the source distribution IIUC. So the only way to have the entrypoint work at all is to checkout bjoern from Git and deploy from there. 

Maybe there’s something that I fundamentally don’t understand about the way you install/deploy bjoern and the web application. Sorry. Please help me understand.

dataflake:
The contrib module referred to in the entry point is the new file at `bjoern/contrib.py`. It's not something external, and nothing needs to be installed. It's being added to the `bjoern` egg  as part of this PR.

You can't import it from code or at the command line when the `bjoern` egg is installed. This is true. But no one needs to do that. setuptools magic makes the entry point work. And that's the only goal here: To make the entry point work. Making the code accessible via import is not a goal.

In my use case (and in everybody else's use case) `bjoern` is installed as an egg. Nothing special.

If you're wondering how entry points work in general, well, I can't tell you. That's deep setuptools voodoo.

dataflake:
Thank you for listening, but I ended up creating a separate shim package so I can continue with my work. Closing this here.

jonashaag:
I’m sorry! I was too busy the last few days. If at some point you can’t be bothered with maintaining a separate package anymore feel free to reopen this.

Issue #162, opened 2019-07-14
Expect/Continue


tjb1982:
from PEP-3333:

> HTTP 1.1 Expect/Continue
> Servers and gateways that implement HTTP 1.1 must provide transparent support for HTTP 1.1's "expect/continue" mechanism. This may be done in any of several ways:
> 
> 1. Respond to requests containing an Expect: 100-continue request with an immediate "100 Continue" response, and proceed normally.
> 2. Proceed with the request normally, but provide the application with a wsgi.input stream that will send the "100 Continue" response if/when the application first attempts to read from the input stream. The read request must then remain blocked until the client responds.
> 3. Wait until the client decides that the server does not support expect/continue, and sends the request body on its own. (This is suboptimal, and is not recommended.)
>
> Note that these behavior restrictions do not apply for HTTP 1.0 requests, or for requests that are not directed to an application object. For more information on HTTP 1.1 Expect/Continue, see RFC 2616, sections 8.2.3 and 10.1.1.

The behavior I'm experiencing in bjoern is the ("suboptimal") third mechanism. Would prefer the second.

PR #163, opened 2019-07-14
#162: naïve implementation of 100-continue

tjb1982:
This is my first Python C extension development experience, and frankly I don't work with C very often, but I thought this would be a fun way to practice.

So if you have time and don't mind reviewing this, I would love to know what could be improved, etc.

I didn't see a good way to implement option 2 in the PEP 3333 suggested mechanisms for handling Expect/Continue, so I went with option 1 (i.e., if Expect is encountered and the value is "100-continue", just go ahead and write it). It would probably be better to allow the client code to make this determination, but, although we could pass the `client_fd` as a PyFile to the client, it doesn't really have access to the header until the entire request is parsed, and by that time, it's too late. Please correct me if I'm wrong.

Thanks!

### DIFF
--- a/bjoern/request.c
+++ b/bjoern/request.c
@@ -147,6 +147,14 @@ on_header_field(http_parser* parser, const char* field, size_t len)
     return 0;
   }
 
+  /* If `Expect` is encountered, set request->state to notify that the client
+   * expects an "HTTP/1.1 100 Continue" response before it will send the body */
+  if (!strncmp(field, HTTP_EXPECT, len)) {
+    if (parser->http_major > 0 && parser->http_minor > 0) {
+      REQUEST->state.expect_continue = true;
+    }
+  }
+
   char field_processed[len];
   for(size_t i=0; i<len; i++) {
     char c = field[i];
--- a/bjoern/request.c
+++ b/bjoern/request.c
@@ -147,14 +147,6 @@ on_header_field(http_parser* parser, const char* field, size_t len)
     return 0;
   }
 
-  /* If `Expect` is encountered, set request->state to notify that the client
-   * expects an "HTTP/1.1 100 Continue" response before it will send the body */
-  if (!strncmp(field, HTTP_EXPECT, len)) {
-    if (parser->http_major > 0 && parser->http_minor > 0) {
-      REQUEST->state.expect_continue = true;
-    }
-  }
-
   char field_processed[len];
   for(size_t i=0; i<len; i++) {
     char c = field[i];
@@ -188,7 +180,49 @@ on_header_value(http_parser* parser, const char* value, size_t len)
   if(!PARSER->invalid_header) {
     /* Set header, or append data to header if this is not the first call */
     _set_or_append_header(REQUEST->headers, PARSER->field, value, len);
+
+    /*
+    ** HTTP/1.1 describes the header `Expect: 100-continue`, which provides
+    ** a mechanism for a client to request permission from a server to send
+    ** the rest of the request (i.e., the body). This restriction is entirely
+    ** client-side in that, regardless of this header existing, the client
+    ** can opt to send the body immediately (i.e., contradicting the header).
+    ** In that case, the header MAY be ignored (according to the RFC). It's
+    ** important to note that this mechanism is not intended to mitigate DoS
+    ** attacks, etc. It's an optional client-side courtesy only.
+    **
+    ** Some popular clients (e.g., cURL) do in fact wait for an indefinite
+    ** time period before sending the rest of the request, and so it is
+    ** appropriate to respond with `HTTP/1.1 100 Continue` in that case.
+    **
+    ** In the future, it may be a reasonable improvement to provide a callback
+    ** API to provide a mechanism for calling code (i.e., Python) to determine
+    ** whether continuation is appropriate based on the existing headers, but
+    ** currently the behavior implemented here is to respond with 100-continue
+    ** automatically as soon as the Expect header is detected, or
+    ** `417 Expectation Failed` if the header contains anything other than
+    ** "100-continue".
+    **
+    ** see https://tools.ietf.org/html/rfc2616#page-48 for specifics.
+    */
+    if (REQUEST->state.expect_continue || REQUEST->state.error_code)
+        return 0;
+
+    if (parser->http_major > 0 && parser->http_minor > 0) {
+      bool field_is_http_expect = !_PEP3333_String_CompareWithASCIIString(
+        PARSER->field, "HTTP_EXPECT"
+      );
+
+      if (field_is_http_expect) {
+        if (!strncmp(value, "100-continue", len)) {
+          REQUEST->state.expect_continue = true;
+        } else {
+          REQUEST->state.error_code = HTTP_EXPECTATION_FAILED;
+        }
+      }
+    }
   }
+
   return 0;
 }
 
@@ -196,6 +230,14 @@ static int
 on_body(http_parser* parser, const char* data, const size_t len)
 {
   PyObject *body;
+
+  /*
+  ** If you're reading the body, you're not waiting for it, so
+  ** no need to respond 100-continue.
+  ** See RFC 2616 https://tools.ietf.org/html/rfc2616#page-49 for details.
+  ** See also comment above in `on_header_value`.
+  */
+  REQUEST->state.expect_continue = false;
 
   body = PyDict_GetItem(REQUEST->headers, _wsgi_input);
   if (body == NULL) {
--- a/bjoern/server.c
+++ b/bjoern/server.c
@@ -224,6 +224,11 @@ ev_io_on_read(struct ev_loop* mainloop, ev_io* watcher, const int events)
         request->current_chunk = _PEP3333_Bytes_FromString(
           http_error_messages[HTTP_SERVER_ERROR]);
       }
+    } else if (request->state.expect_continue) {
+      read_state = done;
+      request->current_chunk = _PEP3333_Bytes_FromString(
+        CONTINUE_RESPONSE);
+      request->state.keep_alive = true;
     } else {
       /* Wait for more data */
       read_state = not_yet_done;
@@ -285,8 +290,16 @@ ev_io_on_write(struct ev_loop* mainloop, ev_io* watcher, const int events)
     if(request->state.keep_alive) {
       DBG_REQ(request, "done, keep-alive");
       ev_io_stop(mainloop, &request->ev_watcher);
-      Request_clean(request);
-      Request_reset(request);
+
+      if (!request->state.expect_continue) {
+        /* There will be more to receive/send after sending 100-Continue, so
+         * don't clobber the request until after the "real" response. */
+        Request_clean(request);
+        Request_reset(request);
+      }
+
+      request->state.expect_continue = false;
+
       ev_io_init(&request->ev_watcher, &ev_io_on_read,
                  request->client_fd, EV_READ);
       ev_io_start(mainloop, &request->ev_watcher);
--- a/bjoern/server.c
+++ b/bjoern/server.c
@@ -29,13 +29,17 @@ static const char* http_error_messages[4] = {
   NULL, /* Error codes start at 1 because 0 means "no error" */
   "HTTP/1.1 400 Bad Request\r\n\r\n",
   "HTTP/1.1 406 Length Required\r\n\r\n",
+  "HTTP/1.1 417 Expectation Failed\r\n\r\n",
   "HTTP/1.1 500 Internal Server Error\r\n\r\n"
 };
+
+static const char *CONTINUE = "HTTP/1.1 100 Continue\r\n\r\n";
 
 enum _rw_state {
   not_yet_done = 1,
   done,
   aborted,
+  expect_continue,
 };
 typedef enum _rw_state read_state;
 typedef enum _rw_state write_state;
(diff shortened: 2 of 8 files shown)


jonashaag:
Thanks for the patch! This looks pretty clean, although I'll have to familiarize myself with HTTP 100 again before I can do a review.

In any case please add at least 1 manual test case, see `tests/` directory for other (manual/automated) tests

jonashaag:
First, minor nitpick: You should check for HTTP/1.1 in the request version, otherwise I believe we can ignore the Expect header.

Not happy with the `write()` call in the parser code. Yes it's the place where we first learn of the Expect header, but it's on the wrong abstraction level. It also circumvents asynchrony/libev entirely, and it also doesn't have error handling. Maybe you can find a way to use the normal write routines. Start looking in `server.c:214`. If you have any questions about the lifecycle of a request in `server.c` feel free to ask.

tjb1982:
Thanks for the fast response! `server.c:214` was where I landed first, actually. I was looking for a place to provide the `client_fd` socket to the `environ`, giving control of this mechanism to the wsgi app, where it seemed more natural to me. The problem with that seems to be that the `request->state.parse_finished` state is only true when the body is also received. That doesn't happen in time for us to respond 100 because the client doesn't send the body until it's done waiting for the continue response (eventually forced to ignore it).

So maybe we could listen for a different state? Is there something like `request->state.header_finished`? If so, we could just respond with 100-continue at that point. Or, perhaps better would be to have no default and give the client code an optional callback with a wsgi app signature, so that it can handle whether or not to reject based on arbitrary criteria on the Python side. E.g.,

    def always_continue(environ, start_response):
        start_response("100 Continue", [])
        return []

    bjoern.run(wsgi_app, "0.0.0.0", 8080, reuse_port=True, on_expect=always_continue)

The PEP seems to think the right way to handle this is to listen for the body to be read, which seems like it would be difficult with this architecture.

Regarding rejection, it looks like 403/405 and 417 are the relevant status codes. 417 specifically in the case of HTTP/1.0. If we were to go the callback route, we can leave that up to client code except in the HTTP/1.0 case, where we can either ignore `Expect` or respond with 417.

It's possible that all of this is much harder than I think because I'm not totally familiar with how libev works.

tjb1982:
So actually I found `on_headers_complete` in the http parser project (it's just not being used by bjoern yet), so let me take another swing at the response by default in server.c using a new `request->state.headers_complete` entry or something akin to that... stay tuned.

tjb1982:
I was able to get the simple case working by wiring up the `on_headers_complete` callback to `request->state.headers_complete` and added a new state called `has_expect_header`. The main piece of that is in server.c like this:

    ...
    } else if (request->state.headers_complete) {
      bool wsgi_ok = true;
      read_state = done;

      if(request->state.parse_finished) {
        /* HTTP parse complete and successful */
        wsgi_ok = wsgi_call_application(request);
      } else {
        if (request->state.has_expect_header) {
          //wsgi_ok = wsgi_call_application(request);

          request->current_chunk = _PEP3333_Bytes_FromString(
            CONTINUE_RESPONSE);
          request->state.keep_alive = true;
        } else {
          read_state = not_yet_done;
        }
      }

      if (!wsgi_ok) {
        /* Response is "HTTP 500 Internal Server Error" */
        DBG_REQ(request, "WSGI app error");
        assert(PyErr_Occurred());
        PyErr_Print();
        assert(!request->state.chunked_response);
        Py_XCLEAR(request->iterator);
        request->current_chunk = _PEP3333_Bytes_FromString(
          http_error_messages[HTTP_SERVER_ERROR]);
      }
    } else {
    ...

I'm just stopping to get your opinion. I sketched out how an `on_expect` handler might work in this commit, but it's not working for some reason having to do with the `io.BytesIO` calls happening in `on_body` in request.c:

```
  PyObject *temp_data = _PEP3333_Bytes_FromStringAndSize(data, len);
  PyObject *tmp = PyObject_CallMethodObjArgs(body, _write, temp_data, NULL);
```

when I uncomment toggle the code in server.c so that it tries to run `on_expect` as a wsgi_app:

          wsgi_ok = wsgi_call_application(request);

          //request->current_chunk = _PEP3333_Bytes_FromString(
          //  CONTINUE_RESPONSE);
          //request->state.keep_alive = true;

`tmp` ends up being null, which means there was some kind of error, but I don't know how to debug it. And since I don't actually know if you care for this design, I thought I'd reach back out before doing any more debugging. Let me know what you think.

Thanks!

jonashaag:
I don’t think we can bjoern make work with actual HTTP 100 callbacks/logic on the Python side. Except for maybe introducing a new API like you’ve done with the on_expect_handler, but I’d rather not properly support HTTP 100 (as in option 2 of PEP3333) than support new API.

Users can deal with the missing explicit HTTP 100 support in various different ways: add logic on a reverse proxy level / change logic on application level / etc.  

So my suggestion is to deal with HTTP 100 entirely on server level, i.e. in `server.c`. Your approach there looks good, though it’s always difficult to make absolute judgement about a design before it’s 100% implemented.

UNLESS you can find a way to decouple body reading from the request parse lifecycle, then we could go with option 2, but I’m not sure it’s worth the effort.

tjb1982:
In the `Expect: 100-continue` case, decoupling reading the body from reading the header is entirely a consequence of the client's behavior. E.g., curl, when the content-length crosses a certain size threshold, will only send the headers (incl. `Expect: 100-continue`) and wait for the server to respond 100-continue, and the http_parser only gets the header bytes from bjoern's `read` call (because that's all there is to read yet), invokes the `on_headers_complete` callback, and then bjoern's `ev_io_on_read` handler has a choice whether to write a response or keep trying to read (blocking).

I.e., the reading itself is all bjoern. The only way I can think of for the server to separate reading the header from the body is to read from the beginning, 1 char at a time until you encounter `\r\n\r\n`. I've done that here in an old project of my own and didn't notice performance problems from it (likely because the headers section is usually relatively small): https://github.com/tjb1982/lal/blob/master/request.c#L44.

That would only be interesting if bjoern were to implement some kind of `on_headers` callback that would allow the client to react to the headers before reading the body. A similar idea as the `on_expect` business, but puts the control server-side in terms of whether or not it reads the request body. To my mind, that would be more useful than `on_expect` which leaves everything to the client, while bjoern potentially drinks from the proverbial firehose.

But coming back to this PR, re: tests, since this is a matter of client behavior, I've been using the `tests/listen.py` test to manually test this with curl. Did you want me to look into reproducing curl's behavior in a python-based client so we have some automation around this, or is `/tests/listen.py` an adequate "manual" test?

tjb1982:
@jonashaag FYI, I think this PR is probably good to go with these small changes. I did have that last question about tests, though. Do you have any thoughts on that?

jonashaag:
Sorry had no time to review so far. Any kind of automated test would be great, even if it’s only comparing curl debug output to some expected strings.

tjb1982:
No problem at all. I added a small change to the python file `bjoern.py` just returning the socket from `listen` (instead of void) which makes dealing with calling `server_run` a little easier. I also started a `requirements.txt` file for keeping track of test automation dependencies. Probably you'll want to start trying to create a small framework for the tests, but I didn't want to go that far in this PR. And of course I added one test using the `requests` lib and plain sockets to ensure this new functionality works.

tjb1982:
One last commit/comment: It occurred to me that I could remove the `on_headers_complete` handler/state, because it isn't strictly necessary. Since the parser "knows" what state it's in, even if the header isn't done being read, we can go ahead and respond to the client to send the body as soon as possible, and while it's doing that we can continue parsing the rest of the header (and eventually the body). It's the tiniest little optimization, but (more importantly) it requires less code and it's more straightforward.

jonashaag:
Not sure if that’s conformant behavior? What if the rest of the header causes a parse error? But it’s probably ok in practise

tjb1982:
That's an interesting question. This is what [RFC 2616](https://tools.ietf.org/html/rfc2616#page-48) says about the use of 100-Continue as it pertains to servers:

> Upon receiving a request which includes an Expect request-header field with the "100-continue" expectation, an origin server MUST either respond with 100 (Continue) status and continue to read from the input stream, or respond with a final status code.

So the appropriate time to respond would depend on how one reads "upon receiving a request."

This is how that RFC defines a request:

        Request       = Request-Line              ; Section 5.1
                        *(( general-header        ; Section 4.5
                         | request-header         ; Section 5.3
                         | entity-header ) CRLF)  ; Section 7.1
                        CRLF
                        [ message-body ]          ; Section 4.3

The request by definition includes both the headers and the body, and so it is incomplete until the body also arrives. The RFC doesn't specify finer grains of incompleteness when it comes to 100-Continue. However, the implication is that the server would respond based on what it finds in the headers. In our case, though, the policy we've come up with is to respond 100-Continue regardless, so we don't have that dependency.

Also, I think we can reasonably assume that a client would wait to recv from the socket until after it is done sending. But even if it doesn't wait, the directive indicates only that the client should continue sending the complete request, including both headers and body.

That said, on the subject of not conforming, this implementation doesn't ensure that the value of the Expect header is "100-Continue." But I did that on purpose: it takes work to check for it, but "100-Continue" is the only valid value, so I thought the existence of the header should be enough, but that's actually not true. We need to respond with 417 if, after checking for "100-Continue", we find something else. I overlooked that part of the spec here:

> The expectation given in an Expect request-header field (see section 14.20) could not be met by this server, or, if the server is a proxy, the server has unambiguous evidence that the request could not be met by the next-hop server.

...

> A server that does not understand or is unable to comply with any of
   the expectation values in the Expect field of a request MUST respond
   with appropriate error status. The server MUST respond with a 417
   (Expectation Failed) status if any of the expectations cannot be met
   or, if there are other problems with the request, some other 4xx
   status.
> This header field is defined with extensible syntax to allow for
   future extensions. If a server receives a request containing an
   Expect field that includes an expectation-extension that it does not
   support, it MUST respond with a 417 (Expectation Failed) status.

I overlooked the fact that this "MUST" be implemented, so let me add a check for "100-Continue" back to this PR too.

tjb1982:
@jonashaag I made some more substantial changes to support the above (417 response, mainly) and added a few more tests. Let me know if you have any concerns.

bjoern/request.c
```diff
@@ -180,7 +180,22 @@ on_header_value(http_parser* parser, const char* value, size_t len)
   if(!PARSER->invalid_header) {
     /* Set header, or append data to header if this is not the first call */
     _set_or_append_header(REQUEST->headers, PARSER->field, value, len);
+
+    if (parser->http_major > 0 && parser->http_minor > 0) {
```
jonashaag: Minor optimisation: We can skip the string comparison if any of `.expect*` is already true

bjoern/server.c
```diff
@@ -213,8 +238,8 @@ ev_io_on_read(struct ev_loop* mainloop, ev_io* watcher, const int events)
     } else if(request->state.parse_finished) {
```
jonashaag: I assume parse finished means we've received header and body? Could you add that to the comment below

bjoern/server.c
```diff
@@ -224,6 +249,12 @@ ev_io_on_read(struct ev_loop* mainloop, ev_io* watcher, const int events)
         request->current_chunk = _PEP3333_Bytes_FromString(
           http_error_messages[HTTP_SERVER_ERROR]);
       }
+    } else if (request->state.expect_continue) {
+      read_state = expect_continue;
+    } else if (request->state.expectation_failed) {
+      read_state = done;
+      request->current_chunk = _PEP3333_Bytes_FromString(
+        EXPECTATION_FAILED);
```
jonashaag: We already have `http_error_messages[...]`, maybe you can move `EXPECTATION_FAILED` there.

bjoern/server.c
```diff
@@ -233,12 +264,18 @@ ev_io_on_read(struct ev_loop* mainloop, ev_io* watcher, const int events)
...
+      CONTINUE_RESPONSE);
+    start_writing(mainloop, request);
+    break;
   case done:
     DBG_REQ(request, "Stop read watcher, start write watcher");
     ev_io_stop(mainloop, &request->ev_watcher);
-    ev_io_init(&request->ev_watcher, &ev_io_on_write,
-               request->client_fd, EV_WRITE);
-    ev_io_start(mainloop, &request->ev_watcher);
+    clear_expectation_state(request);
```
jonashaag: Why do we need this here?

bjoern/server.c
```diff
@@ -278,23 +315,35 @@ ev_io_on_write(struct ev_loop* mainloop, ev_io* watcher, const int events)
     write_state = on_write_chunk(mainloop, request);
   }
 
+  write_state = request->state.expect_continue
```
jonashaag: Could you add a little bit of documentation on HTTP 100 handling to the code (a few sentences should be enough). Maybe to the `ev_io_on_write` docstring? Or where you think it fits.

jonashaag:
VERY good patch. Minor comments see above. Should be ready to merge then, thanks a million!

bjoern/server.c
jonashaag: This = `clear_expectation_state`

tjb1982: There is a condition where `expect_continue` might be `true` but the client decided not to wait (I describe this in the new doc/comment in `on_header_value` in request.c). But I moved this toggle to a more appropriate place (`on_body` in request.c) and it's no longer required here.

tjb1982:
Awesome! I think I addressed all of the above requests and hit some drive-by refactoring along the way, just to preempt a lot of `strlen`s I noticed in wsgi.c. I had a lot of fun with this. I'll look out for some other stuff when I have a moment soon. But do let me know if you have any concerns with this latest volley. Thanks!

jonashaag:
Can I ask you to separate the refactorings to a separate PR? Makes things easier to review and test.

Also I'm not sure what's the point of moving all strings to a constant unless they're used many times and we're at the risk of typos. A level of indirection, however simple, always comes at a cost (= increases complexity/worsens "readability"). `strlen("Constant string")` has no runtime cost. Stuff like `"\r\nfoobar\r\n\r\n" may be a bit verbose but it's very very simple and straightforward to read and understand.

tjb1982:
Oops... I didn't realize there wasn't a runtime cost for `strlen("...")`. That was pretty much the reason I did any of that. I'm still relatively new to C.

jonashaag:
C compilers optimize very well ;)

tjb1982:
Yeah, it was kind of good (selfishly) that I did that because making that mistake caused me to do a little research on compiler optimization so I'll be a little more aware of that in the future.

tjb1982:
FYI, I found a typo I want to remove... one sec.

tjb1982:
It was just an unused variable in the test. Okay, I'm done. Thanks again!

Issue #121, opened 2018-01-24
Wheel support?


Varbin:
Having wheels on PyPi could make installation a lot faster and may also solve the "library headers are required for installation"-problem.

Varbin:
Some CI services could be used. On Travis OS X containers can be used (but I have no experience with them) and AppVeyor  (See https://packaging.python.org/guides/supporting-windows-using-appveyor/) makes it possible  to build on Windows - still this all has to be set up.

For Linux an "manylinux" wheel would be great, it is possible to also build them on travis (example project: https://github.com/pypa/python-manylinux-demo).

jonashaag:
Ok, I'm happy to incorporate this into the release process as long as it means I don't have to install Linux or Windows. Even better if this could be done completely on CI or some trustworthy third party service.

jonashaag:
(Docker is fine as well)

Varbin:
@jonashaag The manylinux wheel creation is done with Docker.

jonashaag:
I see; I'll have a look

jonashaag:
Looks good but seems to be a bit of work. Feel free to contribute a working PR :)

Varbin:
As Continuous Integration could also run a test suite, would you mind if I convert some of the present tests into some independently running automated tests? So before building wheels, it will be tested if nothing broke.

Varbin:
Before doing so, I would like to finish #122/#123 first.

jonashaag:
> As Continuous Integration could also run a test suite, would you mind if I convert some of the present tests into some independently running automated tests? So before building wheels, it will be tested if nothing broke.

That would be absolutely awesome. See also #52. I never got around to doing it.

Jitsusama:
I would like to say that I find this PR quite interesting. Based upon corporate restraints, I have to develop on a Windows machine, even though I deploy to Linux. It would be nice to be able to have bjoern run on my development environment.

dgilman:
I have binary wheels built for several Debian and Ubuntu releases here https://github.com/dgilman/bjoern/releases

bersace:
Does the wheel embed libev or do you still need to install libev4 ? does manylinux requires plateforme to ship libev ?

hoefling:
Here's my humble suggestion: a Travis job configuration that builds fat Linux/MacOS wheels with `libev` included, installs them and runs unit tests, with additional releasing on tag push. Examples:

[...]

PR #164, opened 2019-08-01
Port of test scripts to unit tests

### LINKED ISSUE #52: Turn as many tests into unittests as possible

hoefling:
Closes #52.

### DIFF
--- a/pytests/conftest.py
+++ b/pytests/conftest.py
@@ -1,0 +1,244 @@
+from tblib import pickling_support
+
+pickling_support.install()
+
+import contextlib
+import functools
+import multiprocessing as mp
+import os
+import posixpath
+import signal
+import socket
+import sys
+import tempfile
+from six import reraise as raise_
+from six.moves.urllib.parse import quote, urljoin, urlunsplit
+
+import pytest
+import requests
+import requests_unixsocket
+
+import bjoern
+
+
+def free_port():
+    with contextlib.closing(socket.socket(socket.AF_INET, socket.SOCK_STREAM)) as s:
+        s.bind(('', 0))
+        s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+        port = s.getsockname()[1]
+    return port
+
+
+class ErrorHandleMiddleware:
+    def __init__(self, app, pipe=None):
+        self.app = app
+        self.pipe = pipe
+
+    def __call__(self, *args, **kwargs):
+        try:
+            return self.app(*args, **kwargs)
+        except Exception as e:
+            if self.pipe is not None:
+                exc_info = sys.exc_info()
+                exc_type = exc_info[0]
+                tb = exc_info[2]
+                self.pipe.send((exc_type, e, tb))
+            raise e
+
+
+@pytest.hookimpl(hookwrapper=True)
+def pytest_runtest_call(item):
+    yield
+    if client.__name__ in item.funcargs.keys():
+        # check for uncaught errors in server process, raise if any
+        testclient = item.funcargs[client.__name__]
+        testclient._reraise_app_errors()
+
+
+def reraise(f):
+    @functools.wraps(f)
+    def wrapper(*args, **kwargs):
+        result = f(*args, **kwargs)
+        testclient = args[0]
+        testclient._reraise_app_errors()
+        return result
+
+    return wrapper
+
+
+class Client(object):
+    host = None
+    port = None
+
+    _error_publisher = None
+    _error_receiver = None
+    _proc = None
+    _session = None
+    _sock = None
+
+    def __init__(self):
+        self.reuse_port = False
+        self.listen_backlog = bjoern.DEFAULT_LISTEN_BACKLOG
+        self._children = []
+
+    @property
+    def session(self):
+        return self._session
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, type, value, tb):
+        self.stop()
+
+    def start(self, wsgi_app, num_workers=1):
+        assert self._sock is not None
+        num_workers = max(num_workers, 1)
+        (self._error_receiver, self._error_publisher) = mp.Pipe(False)
+        for _ in range(num_workers):
+            errorhandler = ErrorHandleMiddleware(wsgi_app, self._error_publisher)
+            child = mp.Process(
+                target=bjoern.server_run, args=(self._sock, errorhandler)
+            )
+            child.start()
+            self._children.append(child)
+
+    def stop(self):
+        if self._sock is None:
+            return
+        self.session.close()
+        self._error_publisher.close()
+        self._error_receiver.close()
+        for child in self._children:
+            os.kill(child.pid, signal.SIGTERM)
+            child.join()
+        if self._sock.family == socket.AF_UNIX:
+            filename = self._sock.getsockname()
+            if filename[0] != '\0':
+                os.unlink(self._sock.getsockname())
+        self._sock.close()
+
+    def _reraise_app_errors(self, timeout=0):
+        exc_info = None
+        try:
+            if self._error_receiver.poll(timeout):
+                exc_info = self._error_receiver.recv()
+        except EOFError:
+            return
+        if exc_info is not None:
+            raise_(*exc_info)  # python2 compat
+
+    @reraise
+    def get(self, path=None, **kwargs):
+        """Adapter for :py:meth:`requests.Session.get`"""
+        return self.session.get(self._join(self.root, (path or '')), **kwargs)
+
+    @reraise
+    def options(self, path=None, **kwargs):
+        """Adapter for :py:meth:`requests.Session.options`"""
+        return self.session.options(self._join(self.root, (path or '')), **kwargs)
+
+    @reraise
+    def head(self, path=None, **kwargs):
+        """Adapter for :py:meth:`requests.Session.head`"""
+        return self.session.head(self._join(self.root, (path or '')), **kwargs)
+
+    @reraise
+    def post(self, path=None, data=None, json=None, **kwargs):
+        """Adapter for :py:meth:`requests.Session.post`"""
+        return self.session.post(
+            self._join(self.root, (path or '')), data=data, json=json, **kwargs
+        )
+
+    @reraise
+    def put(self, path=None, data=None, **kwargs):
+        """Adapter for :py:meth:`requests.Session.put`"""
+        return self.session.put(
+            self._join(self.root, (path or '')), data=data, **kwargs
+        )
+
+    @reraise
+    def patch(self, path=None, data=None, **kwargs):
+        """Adapter for :py:meth:`requests.Session.patch`"""
+        return self.session.patch(
+            self._join(self.root, (path or '')), data=data, **kwargs
+        )
+
+    @reraise
+    def delete(self, path=None, **kwargs):
+        """Adapter for :py:meth:`requests.Session.delete`"""
+        return self.session.delete(self._join(self.root, (path or '')), **kwargs)
+
+
+class HttpClient(Client):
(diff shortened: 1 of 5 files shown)


pytests/conftest.py
```diff
@@ -0,0 +1,244 @@
...
+        # check for uncaught errors in server process, raise if any
+        testclient = item.funcargs[client.__name__]
+        testclient._reraise_app_errors()
+
+
+def reraise(f):
+    @functools.wraps(f)
+    def wrapper(*args, **kwargs):
+        result = f(*args, **kwargs)
+        testclient = args[0]
```
jonashaag: better make this the first `wrapper` argument explicitly

pytests/conftest.py
```diff
@@ -0,0 +1,244 @@
...
+
+    def __call__(self, *args, **kwargs):
+        try:
+            return self.app(*args, **kwargs)
+        except Exception as e:
+            if self.pipe is not None:
+                exc_info = sys.exc_info()
+                exc_type = exc_info[0]
+                tb = exc_info[2]
+                self.pipe.send((exc_type, e, tb))
```
jonashaag: why not send entire `exc_info`

pytests/conftest.py
```diff
@@ -0,0 +1,244 @@
...
+import requests
+import requests_unixsocket
+
+import bjoern
+
+
+def free_port():
+    with contextlib.closing(socket.socket(socket.AF_INET, socket.SOCK_STREAM)) as s:
+        s.bind(('', 0))
+        s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
```
jonashaag: I guess this might break on systems that don't support reuseaddr?

pytests/conftest.py
```diff
@@ -0,0 +1,244 @@
...
+        try:
+            if self._error_receiver.poll(timeout):
+                exc_info = self._error_receiver.recv()
+        except EOFError:
+            return
+        if exc_info is not None:
+            raise_(*exc_info)  # python2 compat
+
+    @reraise
+    def get(self, path=None, **kwargs):
```
jonashaag: Maybe refactor these into a single method and use `functools.partial`? But that's nitpicking

pytests/conftest.py
```diff
@@ -0,0 +1,244 @@
...
+
+@pytest.fixture
+def unixclient():
+    """
+    * don't use pytest's tmpdir/tmp_path as the generated paths easily exceed 108 chars,
+    causing an "OSError: AF_UNIX path too long" on MacOS
+    * also, filenames longer than 64 chars cause urllib hickups, see https://bugs.python.org/issue32958
+    * another thing: tempfile.gettempdir() on MacOS resorts to $TMPDIR
+    which is autogenerated into /var/folders and is also longer than 64 chars
+    """
```
jonashaag: btw there's also anonymous sockets with `@...`, but not supported by all platforms. maybe we can test these as well.

pytests/test_bjoern.py
```diff
@@ -0,0 +1,464 @@
...
+@pytest.mark.parametrize('header_x_auth', ('X-Auth_User', 'X_Auth_User', 'X_Auth-User'))
+def test_CVE_2015_0219(client, header_x_auth):
+    """
+    https://www.djangoproject.com/weblog/2015/jan/13/security/
+    https://snyk.io/vuln/SNYK-PYTHON-BJOERN-40507
+    """
+
+    def app(env, start_response):
+        assert header_x_auth not in env.keys()
+        assert 'X-Auth-User' not in env.keys()
```
jonashaag: Shouldn't these headers be transformed to `HTTP_X_AUTH_USER` etc (since WSGI servers do that before putting them to env)

pytests/test_bjoern.py
```diff
@@ -0,0 +1,464 @@
...
+
+def wrap(text, width=20, placeholder='...'):
+    wrapped = textwrap.wrap(text, width=width)[0]
+    if len(wrapped) < len(text):
+        wrapped += placeholder
+    return wrapped
+
+
+@pytest.mark.parametrize(
+    'header_key,header_value',
```
jonashaag: This decorator is very difficult to read, maybe you can reformat

pytests/test_bjoern.py
```diff
@@ -0,0 +1,464 @@
...
+    def app(env, start_response):
+        f = PseudoFile() if pseudo else open(file, 'rb')
+        wrapped = wrapper(f, env)
+        start_response('200 ok', [('Content-Length', str(os.path.getsize(file)))])
+        return wrapped
+
+    return app
+
+
+@pytest.mark.parametrize('file', ('README.rst',), ids=('README.rst',))
```
jonashaag: Better use lists here, `(foo,)` is difficult to parse

pytests/test_bjoern.py
```diff
@@ -0,0 +1,464 @@
...
+@pytest.mark.parametrize(
+    'pseudo', (False, True), ids=lambda x: 'pseudofile' if x else 'file'
+)
+@pytest.mark.parametrize(
+    'wrapper',
+    (
```
jonashaag: Better use `[]` here as well.

pytests/test_bjoern.py
```diff
@@ -0,0 +1,464 @@
...
+        lambda f, _: iter(lambda: f.read(64 * 1024), b''),
+        lambda f, _: f,
+        lambda f, env: env['wsgi.file_wrapper'](f),
+        lambda f, env: env['wsgi.file_wrapper'](f, 1),
+    ),
+    ids=('callable-iterator', 'xreadlines', 'filewrapper', 'filewrapper2'),
```
jonashaag: And here and in any place where an actual list of things is provided ;)

pytests/test_bjoern.py
```diff
@@ -0,0 +1,464 @@
...
+    def _app(environ, start_response):
+        start_response('200 OK', [('Content-Type', 'text/plain')])
+        return [b'Hello World']
+
+    client.start(_app)
+    client.get()
+
+
+@pytest.mark.parametrize(
+    'expect_header_value,content_length,body,response',
```
jonashaag: This is also difficult to parse

pytests/test_bjoern.py
```diff
@@ -0,0 +1,464 @@
...
+    # use the raw connection instead of httpclient
+    with contextlib.closing(
+        socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    ) as raw_client:
+        raw_client.connect((httpclient.host, httpclient.port))
+        request = 'GET /fizz HTTP/1.1\r\nExpect: {}\r\nContent-Length: {}\r\n\r\n{}'.format(
+            expect_header_value, content_length, body
+        )
+        raw_client.send(request.encode('utf-8'))
+        resp = raw_client.recv(1 << 10)
```
jonashaag: Why not simply write 1024 here

jonashaag:
This is really great. Thanks a million.

Minor suggestions in the review.

Maybe you can also add some proper test descriptions rather than `tests/fileabc.py` -- at this point you know better than I what the tests do! (Single sentence should be enough!)

One thing we have to think about is that some of the tests are useful only when called 1000s of times, for example the exception tests, which are used to check for memory leaks. Not sure if that's easily automated; otherwise there should be a way to run a list of WSGI applications without the actual test code so we can use the "old" way of bombarding bjoern using `make ab` etc

pytests/conftest.py
hoefling: Indeed, `socket.REUSEADDR` should be handled same as `socket.REUSEPORT`.

pytests/conftest.py
hoefling: Agreed, good catch!

pytests/conftest.py
hoefling: Agreed!

pytests/conftest.py
hoefling: No, I agree on that, good suggestion. I have mainly added the method copies because of code completion in IDE, but typing stubs would be way better here.

pytests/conftest.py
hoefling: Good idea! Unfortunately, it looks like `urllib` can't parse the URL, so `requests_unixsocket` can't connect - although it states [the support for abstract unix sockets](https://github.com/msabramo/requests-unixsocket#abstract-namespace-sockets). I have left [a comment with the error message](https://github.com/msabramo/requests-unixsocket/issues/18#issuecomment-517752129), waiting for the author's response. Maybe extract it in a separate task?

pytests/test_bjoern.py
hoefling: Oh man, this is so embarrassing. Thanks for spotting this!

pytests/test_bjoern.py
hoefling: This comes from the `black` autoformatter which runs on every commit on my machine. I will try configure `black` to use a bigger line length argument, the formatting is indeed questionable...

pytests/test_bjoern.py
hoefling: Agreed!

pytests/test_bjoern.py
hoefling: This is due to code copying from `tests/expect100.py`. Shouldn't it even be `bjoern.DEFAULT_LISTEN_BACKLOG`?

hoefling:
> One thing we have to think about is that some of the tests are useful only when called 1000s of times, for example the exception tests, which are used to check for memory leaks.

I will look into it, however I'm not sure whether it is a task for unit tests and `pytest`. Does it look more like load testing? Usually I write load tests with [`locust`](https://locust.io/), would it be an option for you?

pytests/test_bjoern.py
tjb1982: The reason I did that in expect100.py (I'm the author of that specific test) is because I consider it more legible when dealing with byte counts between KB -> MB -> GB. E.g.,

KB = 1024
MB = 1048576
GB = 1073741824

vs.

KB = 1 << 10
MB = 1 << 20
GB = 1 << 30

or, if your test grows the content-length exponentially with each iteration, such as that test does [here](https://github.com/jonashaag/bjoern/blob/master/tests/expect100.py#L45)

jonashaag:
> I will look into it, however I'm not sure whether it is a task for unit tests and `pytest`. Does it look more like load testing? Usually I write load tests with [`locust`](https://locust.io/), would it be an option for you?

Sorry for the delay!

Locust is fine for me, or any other tool pretty much.

Yes it's more like load testing. Best case we can load test a bjoern instance and monitor memory use

jonashaag:
BUT some of the tests I also used for manual load/performance testing, for example when comparing requests/s for two revisions of bjoern.

gcavalcante8808:
Is there any plans to merge this PR? Its great for helping development.

jonashaag:
Yes when the things I mentioned above are implemented.

mamico:
I think that an idea could be to merge this PR and add an issue (improvement) for missing tests, probably, if @hoefling are currently busy, other contributors could help on. @jonashaag what do you think about?

jonashaag:
No. That does not solve the problem that nobody is working on this PR and at the moment.

hoefling:
This PR (and the other one I have started working at for #121) were parts of making `bjoern` pass the requirements of our internal QA & security team. Since then, we have moved from WSGI, so `bjoern` is not a requirement anymore; I thus don't have the time to work on the tests anymore. Feel free to use the code in this PR if you want to finish the task - just create your own PR and apply the diff from this one.

However, I don't see any issues with merging the current state as IIRC the existing test scripts are not removed so they are still usable for manual load tests.

jonashaag:
You have a policy that 3rd party applications must have an automated test suite (of any quality)? 😀

PR #168, opened 2019-10-01
Add statsd metrics on server events

### LINKED ISSUE #167: Support for telemetry
Is there any plan to expose metrics? Something like statsd or prometheus sink should work for most people.
I can work on statsd or dogstatsd support if it hasn't been ruled out already.

Gufran:
I've added metrics wherever they made sense to me, these are all simple `increment` metrics counting the number of times an event happened.

There is scope for `gauge` and `timer` too but I'm not sure those will be as helpful inside bjoern, almost all of these can also be exposed from the WSGI app.

I don't write C a lot so please feel free to suggest code design changes.

closes #167

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,11 +9,15 @@ HTTP_PARSER_DIR	= http-parser
 HTTP_PARSER_OBJ = $(HTTP_PARSER_DIR)/http_parser.o
 HTTP_PARSER_SRC = $(HTTP_PARSER_DIR)/http_parser.c
 
-objects		= $(HTTP_PARSER_OBJ) \
+STATSD_CLIENT_DIR = statsd-c-client
+STATSD_CLIENT_OBJ = $(STATSD_CLIENT_DIR)/statsd-client.o
+STATSD_CLIENT_SRC = $(STATSD_CLIENT_DIR)/statsd-client.c
+
+objects		= $(HTTP_PARSER_OBJ) $(STATSD_CLIENT_OBJ) \
 		  $(patsubst $(SOURCE_DIR)/%.c, $(BUILD_DIR)/%.o, \
 		             $(wildcard $(SOURCE_DIR)/*.c))
 
-CPPFLAGS	+= $(PYTHON_INCLUDE) -I . -I $(SOURCE_DIR) -I $(HTTP_PARSER_DIR)
+CPPFLAGS	+= $(PYTHON_INCLUDE) -I . -I $(SOURCE_DIR) -I $(HTTP_PARSER_DIR) -I $(STATSD_CLIENT_DIR)
 CFLAGS		+= $(FEATURES) -std=c99 -fno-strict-aliasing -fcommon -fPIC -Wall
 LDFLAGS		+= $(PYTHON_LDFLAGS) -l ev -shared -fcommon
 
@@ -107,3 +111,6 @@ upload:
 
 $(HTTP_PARSER_OBJ):
 	$(MAKE) -C $(HTTP_PARSER_DIR) http_parser.o CFLAGS_DEBUG_EXTRA=-fPIC CFLAGS_FAST_EXTRA=-fPIC
+
+$(STATSD_CLIENT_OBJ):
+	$(MAKE) -C $(STATSD_CLIENT_DIR) statsd-client.o CFLAGS_DEBUG_EXTRA=-fPIC CFLAGS_FAST_EXTRA=-fPIC
--- a/Makefile
+++ b/Makefile
@@ -21,10 +21,6 @@ CPPFLAGS	+= $(PYTHON_INCLUDE) -I . -I $(SOURCE_DIR) -I $(HTTP_PARSER_DIR) -I $(S
 CFLAGS		+= $(FEATURES) -std=c99 -fno-strict-aliasing -fcommon -fPIC -Wall
 LDFLAGS		+= $(PYTHON_LDFLAGS) -l ev -shared -fcommon
 
-ifneq ($(WANT_SENDFILE), no)
-FEATURES	+= -D WANT_SENDFILE
-endif
-
 ifneq ($(WANT_SIGINT_HANDLING), no)
 FEATURES	+= -D WANT_SIGINT_HANDLING
 endif
--- a/Makefile
+++ b/Makefile
@@ -41,6 +41,10 @@ ifneq ($(WANT_STATSD), no)
 FEATURES	+= -D WANT_STATSD
 endif
 
+ifneq ($(WANT_STATSD_TAGS), no)
+FEATURES	+= -D WANT_STATSD_TAGS
+endif
+
 all: prepare-build $(objects) _bjoernmodule
 
 print-env:
--- a/Makefile
+++ b/Makefile
@@ -37,11 +37,11 @@ ifndef SIGNAL_CHECK_INTERVAL
 FEATURES	+= -D SIGNAL_CHECK_INTERVAL=0.1
 endif
 
-ifneq ($(WANT_STATSD), no)
+ifeq ($(WANT_STATSD), yes)
 FEATURES	+= -D WANT_STATSD
 endif
 
-ifneq ($(WANT_STATSD_TAGS), no)
+ifeq ($(WANT_STATSD_TAGS), yes)
 FEATURES	+= -D WANT_STATSD_TAGS
 endif
 
--- a/Makefile
+++ b/Makefile
@@ -37,6 +37,10 @@ ifndef SIGNAL_CHECK_INTERVAL
 FEATURES	+= -D SIGNAL_CHECK_INTERVAL=0.1
 endif
 
+ifneq ($(WANT_STATSD), no)
+FEATURES	+= -D WANT_STATSD
+endif
+
 all: prepare-build $(objects) _bjoernmodule
 
 print-env:
--- a/Makefile
+++ b/Makefile
@@ -63,7 +63,7 @@ _bjoernmodule:
 again: clean all
 
 debug:
-	CFLAGS='-D DEBUG -g' make again
+	CFLAGS='${CFLAGS} -D DEBUG -g' make again
 
 $(BUILD_DIR)/%.o: $(SOURCE_DIR)/%.c
 	@echo ' -> ' $(CC) $(CPPFLAGS) $(CFLAGS) -c $< -o $@
--- a/README.rst
+++ b/README.rst
@@ -42,16 +42,24 @@ Usage
    # Bind to abstract Unix socket: (Linux only)
    bjoern.run(wsgi_application, 'unix:@socket_name')
 
+   # enable statsd metrics
+   bjoern.run(wsgi_application, host, port, reuse_port=True, enable_statsd=True)
+
 Alternatively, the mainloop can be run separately::
 
    bjoern.listen(wsgi_application, host, port)
    bjoern.run()
-   
+
+   # with metrics
+   bjoern.listen(wsgi_application, host, port)
+   bjoern.run(enable_statsd=True)
+
 You can also simply pass a Python socket(-like) object. Note that you are responsible
 for initializing and cleaning up the socket in that case. ::
 
    bjoern.server_run(socket_object, wsgi_application)
    bjoern.server_run(filedescriptor_as_integer, wsgi_application)
+   bjoern.server_run(socket_object, wsgi_application, enable_statsd=True)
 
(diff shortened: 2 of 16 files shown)


jonashaag:
Thanks! This is great work, and great quality code!

You'll have to help me a bit here. Is statsd a quasi standard of these kinds of metrics? What alternatives exist?

As for the code: We should make the statsd integration optional. You're the first person to request this integration and while I guess it makes sense to have it I'm not sure we should have it by default. Maybe we can simply use compile directives. Or maybe we turn it into a generic metrics interface that we always ship and that you can plug in your own metrics implementation. (But maybe that's overkill architecture wise...)

Gufran:
[statsd as protocol](https://github.com/statsd/statsd#usage) is pretty much the standard, there is an improvised version of statsd [from datadog](https://docs.datadoghq.com/developers/dogstatsd/) that supports tags and is generally preferred over statsd.

Next in line is [Prometheus](https://prometheus.io/docs/introduction/overview/) and it brings its own protocol that is richer and more complex than what statsd can do.

After that you have hosted services that support either statsd or prometheus protocol or people roll out their own solutions with various time series databases available out there.

If you want to go for the standard then I think dogstatsd (statsd from datadog) should be the choice.

> We should make the statsd integration optional.

Sure, I think compiler directives should work. I don't think I can implement the pluggable interface in a reasonable amount of time.

Gufran:
I've made statsd an optional feature. you may want to look at [argument parsing](https://github.com/jonashaag/bjoern/pull/168/files#diff-62837363484d57f159a8d9150b2b2e84R23-R28) and let me know if there is a better way to ignore extra arguments in `PyArg_ParseTuple`.

Makefile
```diff
@@ -33,6 +37,10 @@ ifndef SIGNAL_CHECK_INTERVAL
 FEATURES	+= -D SIGNAL_CHECK_INTERVAL=0.1
 endif
 
+ifneq ($(WANT_STATSD), no)
```
jonashaag: Let's disable by default.

README.rst
```diff
@@ -42,16 +42,24 @@ Usage
    # Bind to abstract Unix socket: (Linux only)
    bjoern.run(wsgi_application, 'unix:@socket_name')
 
+   # enable statsd metrics
```
jonashaag: Add comment that this needs manual compilation

bjoern.py
```diff
@@ -70,6 +86,8 @@ def run(*args, **kwargs):
     """
     global _default_instance
 
+    kwargs, statsd_config = __separate_config(kwargs)
```
jonashaag: Why not make the statsd config param as follows

```
def run(*args, **kwargs):
    statsd = kwargs.pop('statsd', None)
```

```
run(statsd={
    'host': ...,
    'port': ....,
})
```

If `statsd` not passed -> disabled.

Type checks in the `.run()` method.

`statsd_` params in `server_run` remain as-is but with `None` defaults.

bjoern/_bjoernmodule.c
```diff
@@ -2,6 +2,7 @@
 #include "server.h"
 #include "wsgi.h"
 #include "filewrapper.h"
+#include "statsd-client.h"
```
jonashaag: This must be wrapped in a compile directive

bjoern/_bjoernmodule.c
```diff
@@ -10,9 +11,22 @@ run(PyObject* self, PyObject* args)
 
   PyObject* socket;
 
-  if(!PyArg_ParseTuple(args, "OO:server_run", &socket, &info.wsgi_app)) {
+#ifdef WANT_STATSD
+  StatsdInfo statsd_info;
```
jonashaag: Maybe we also need a thread identifier in the statsd info since host + port aren't guaranteed to be unique

bjoern/common.h
```diff
@@ -57,4 +57,11 @@ PyObject *_REMOTE_ADDR, *_PATH_INFO, *_QUERY_STRING, *_REQUEST_METHOD, *_GET,
   #define assert(...) do{}while(0)
 #endif
 
+#ifdef WANT_STATSD
+  #include "statsd-client.h"
+  #define STATSD_INC(name) statsd_inc(((ThreadInfo*) ev_userdata(mainloop))->statsd, name, 1.0)
```
jonashaag: Let's call this "increment" or "incr" since "inc" may also mean "include"

bjoern/server.c
```diff
@@ -47,6 +54,9 @@ typedef enum _rw_state write_state;
 typedef struct {
   ServerInfo* server_info;
   ev_io accept_watcher;
+#ifdef WANT_STATSD
+  statsd_link* statsd;
```
jonashaag: Couldn't we put this into `ServerInfo` struct?

bjoern/_bjoernmodule.c
jonashaag: Let's make this local variables with `statsd_` prefix, not a struct, since it's only used temporarily in this routine.

setup.py
```diff
@@ -2,17 +2,19 @@
...
 bjoern_extension = Extension(
     '_bjoern',
     sources       = SOURCE_FILES,
     libraries     = ['ev'],
-    include_dirs  = ['http-parser', '/usr/include/libev', '/opt/local/include'],
+    include_dirs  = ['http-parser', 'statsd-c-client', '/usr/include/libev', '/opt/local/include'],
     define_macros = [('WANT_SENDFILE', '1'),
                      ('WANT_SIGINT_HANDLING', '1'),
                      ('WANT_SIGNAL_HANDLING', '1'),
+                     ('WANT_STATSD', '1'),
```
jonashaag: Disable by default

setup.py
```diff
@@ -2,17 +2,19 @@
 import glob
 from setuptools import setup, Extension
 
-SOURCE_FILES = [os.path.join('http-parser', 'http_parser.c')] + \
+SOURCE_FILES = [os.path.join('http-parser', 'http_parser.c'),
+                os.path.join('statsd-c-client', 'statsd-client.c')] + \
```
jonashaag: Don't ship

Gufran:
@jonashaag thanks for the review. meanwhile I went ahead and added support for tags since tags are way more powerful than simple metrics.
I'll address your comments in a bit.

Gufran:
@jonashaag all reviews addressed.

bjoern/statsd_tags.c
```diff
@@ -0,0 +1,32 @@
+#include <stdio.h>
+#include "statsd-client.h"
+
+#define MAX_MSG_LEN_WITH_TAGS 1000
+
+/* will change the original string */
+static void cleanup(char *stat)
```
jonashaag: Why do we need this? Why don't we simply not put any of these chars in the strings in the source code?

Gufran: yeah, that makes sense. removed the function.

jonashaag:
Looks very good now! Thanks!

jonashaag:
Did some cleanup and removed default config from bjoern (we shouldn't have to deal with statsd config values in bjoern)

Can you retest? Also please test setup.py compilatin method

bjoern.py
```diff
@@ -35,11 +35,13 @@ def bind_and_listen(host, port=None, reuse_port=False,
     return sock
 
 
-def server_run(sock, wsgi_app):
-    _bjoern.server_run(sock, wsgi_app)
+def server_run(sock, wsgi_app, statsd=None):
+    if statsd is None:
+        _bjoern.server_run(sock, wsgi_app)
```
Gufran: this will break when compiled with statsd and statsd options are not provided.

bjoern.py
```diff
@@ -35,11 +35,13 @@ def bind_and_listen(host, port=None, reuse_port=False,
...
+    else:
+        _bjoern.server_run(sock, wsgi_app, int(statsd['enable']), statsd['host'], statsd['port'], statsd['ns'], statsd.get('tags'))
```
Gufran: it forces users to declare all statsd options even if they want to use the defaults. Is this intended?

bjoern/_bjoernmodule.c
```diff
@@ -3,16 +3,42 @@
...
+#else
+  char* ignored_str = NULL;
+  int ignored_int = 0;
+
+  if(!PyArg_ParseTuple(args, "OO|izizz:server_run", &socket, &info.wsgi_app, &ignored_int,
+                       &ignored_str, &ignored_int, &ignored_str, &ignored_str)) {
     return NULL;
   }
+  if (ignored_str != NULL || ignored_int != 0) {
+    PyErr_Format(PyExc_TypeError, "Unexpected statsd_* arguments (forgot to compile with statsd support?)");
```
Gufran: This can be printed as a warning and bjoern can still boot. Ignoring these arguments when statsd support is not compiled will make it easier to deploy different versions of bjoern without having to make changes in code.

instrumentation.md
```diff
@@ -0,0 +1,49 @@
...
+```
+
+After building with statsd support you can enable statsd metrics on server boot by setting `enable_statsd=True`, e.g.
+
+```python
+bjoern.listen(wsgi_application, host, port)
+bjoern.run(statsd={'enable': True, 'host': '...', port: ..., ns: 'bjoern'})
+
+# With tags
+bjoern.run(statsd={'enable': True, 'host': '...', port: ..., ns: 'bjoern', 'tags': 'app:my-app-name,zone:central-europe'})
```
Gufran: I think it is worth clarifying that `tags` is expected to be a comma separated string and not a list of tags. Applications generally expect a list of string so people might struggle here.

Gufran:
@jonashaag just went through it and left a couple of comments. will test the changes in a bit.

Gufran:
Ok, just tested and the branch is broken.

|     Feature/Test     | Test Hello | Test Hello with statsd | Test Hello with statsd and tags |
|-----------|------------|------------------------|---------------------------------|
| without statsd and without statsd tags | Works | Broken. Wont boot with statsd options | Broken. Wont boot with statsd options |
| with statsd and without statsd tags | Broken. Wants statsd options | Works | Works |
| with statsd and with statsd tags | Broken. Wants statsd options | Works | Works |

@jonashaag do you want to make the changes or should I?

bjoern/_bjoernmodule.c
jonashaag: Explicit is better than implicit.

Ignoring arguments (even with warning) is always a terrible idea. Better be explicit about it and throw an error. If you have multiple different versions of bjoern running then this is complexity in your code, not in bjoern, so it should also be dealt with in your code.

bjoern.py
jonashaag: Yes, I don't want any statsd default options inside the bjoern code. These values are completely unrelated to bjoern. For example if the statsd defaults change in future (e.g. they change to a different port) then we shouldn't have to make any changes to bjoern.

Also, it's just 4 values, and it forces the caller code to be explicit rather than implicit, which is always better :)

bjoern.py
jonashaag: Yes, unfortunately in the current implementation we can't know whether bjoern is compiled with statsd support. Maybe we should simply add an attribute that we can check.

instrumentation.md
jonashaag: From what I've seen on the internet the tags format isn't standardised. I've seen versions like `"foo,bar,spam"` and versions like `"foo:x,bar:y,spam:z"`.

Agree with adding a comment!

bjoern.py
Gufran: Sure, I think we should add attributes for all features that can be consulted from python and that way people will be able to write code that can selectively enable or disable a feature.

something like 

 - `bjoern.has_statsd`
 - `bjoern.has_statsd_tags`
 - `bjoern.has_signal_handling`

and so on.

jonashaag: Sounds good. Maybe we can use a single attribute "features" or "compilation flags" or whatever, and make it a list that we simply populate as follows

```
l = PyList_New();
#ifdef WANT_X
PyList_Append(l, PyString_FromString("X"));
#endif 
...
```

Gufran:
@jonashaag take a look now.

jonashaag:
This looks very good now. Did some minor cleanup + debugging. Will merge

Gufran:
@jonashaag any ETA on when will this be available in PyPI? I see it hasn't been tagged yet and the latest version available in PyPI is `3.0.1`.

blaggacao:
Any plans in the near future to follow up with prometheus instrumatation support?

PR #169, opened 2019-10-16
Fixed client accepting flow

BlackJohnny:
I am currently working on a platform that uses the microservice architecture and I am testing the use of bjoern as http server for our microservices. Our microservices are behind envoy proxy and under high stress I get timeouts for about 5% of the connections.

It seems to be a bjoern issue with the "keep-alive" flow.

**Current behavior in bjoern:**

- From time to time, after a request is served on a keep-alive connection, the new request triggers the `accept_watcher/ev_io_on_request` which puts bjoern in "accept mode", therefore the previous connection is terminated.

**Implemented behavior:**

-  After a new client is accepted, the `accept_watcher` is disabled. The watcher is enabled again only on the `close_connection` use case.

This has been tested with the version 3.0.1.

### DIFF
--- a/bjoern/server.c
+++ b/bjoern/server.c
@@ -179,6 +179,8 @@ ev_io_on_request(struct ev_loop* mainloop, ev_io* watcher, const int events)
   DBG_REQ(request, "Accepted client %s:%d on fd %d",
           inet_ntoa(sockaddr.sin_addr), ntohs(sockaddr.sin_port), client_fd);
 
+  /* Stop the accept watcher */
+  ev_io_stop(mainloop, &((ThreadInfo*)ev_userdata(mainloop))->accept_watcher);
   ev_io_init(&request->ev_watcher, &ev_io_on_read,
              client_fd, EV_READ);
   ev_io_start(mainloop, &request->ev_watcher);
@@ -535,4 +537,9 @@ close_connection(struct ev_loop *mainloop, Request* request)
   ev_io_stop(mainloop, &request->ev_watcher);
   close(request->client_fd);
   Request_free(request);
+
+  /* Resume accepting new clients */
+  ThreadInfo* thread_info = (ThreadInfo*)ev_userdata(mainloop);
+  ev_io_init(&thread_info->accept_watcher, ev_io_on_request, thread_info->server_info->sockfd, EV_READ);
+  ev_io_start(mainloop, &thread_info->accept_watcher);
 }
--- a/tests/client_for_server_with_workers.py
+++ b/tests/client_for_server_with_workers.py
@@ -1,0 +1,48 @@
+import logging
+import requests
+import time
+import timeit
+import sys
+import signal
+
+sleep_time = 2
+port = int(sys.argv[1]) if len(sys.argv) == 2 else 8081
+host = f"http://localhost:{port}/"
+
+print(f"Making requests on {host}")
+
+s = requests.Session()
+number_of_calls = 10
+server_pids = {}
+
+def shutdown(sig, frame):
+    logging.warning(f"Requests {server_pids}")
+    sys.exit()
+
+signal.signal(signal.SIGINT, shutdown)
+
+request_no = 0
+error_request_no = 0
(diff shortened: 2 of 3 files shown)


BlackJohnny:
Extra detail on when the issue occurs 100%:

- I start bjoern for N workers with fork
- I start N clients that are constantly issuing requests and they are reusing their respective connections
- I start client N+1. It gets a connection but one of the other clients gets stucked.

5% happens when I have a number of clients less than N.

jonashaag:
Hi thanks for the patch I will have a look soon. If you can reproduce it 100% do you think you could also come up with a „unit test“ that reproduces the issue?

BlackJohnny:
bjoern

BlackJohnny:
In the image above, the left side i have official bjoern build that gives timeouts, therefore the max duration is high compared with with the min duration. On the right side i have bjoern with the updated code.

Maybe another solution would be to explicitly terminate the client connection to allow the connected peer to recycle the connection on its side. In our case, `close_connection` on bjoern side occurs when envoy detects the timeout and closes the connection with bjoern.

Any other suggestions are welcomed :)

BlackJohnny:
PS: the absolute min values are basically when i have only one client instance connected. Once I start the other ones the real min value with concurrent clients is not captured.

jonashaag:
Thanks a lot!

Tests were helpful.

Your patch was incorrect since it essentially made bjoern synchronous (can't have more clients connected than forks).

Problem was that sometimes `accept()` would block (not sure why!). Forgot to set socket to non-blocking.

BlackJohnny:
Thank you for your time!
> Thanks a lot!
> 
> Tests were helpful.
> 
> Your patch was incorrect since it essentially made bjoern synchronous

This is what bjoern does by default because it is single threaded. I think I did not change the current behavior with that fix. **Please confirm.**

> (can't have more clients connected than forks).

This is only true with keep-alive connections. But on a high level isn't this a fact for bjoern? Being a single threaded architecture it will only handle the requests sequentially per worker. Therefore, the concurrent handled clients equals the number of forks. **Please confirm.**

> Problem was that sometimes `accept()` would block (not sure why!). Forgot to set socket to non-blocking.

What I expect from this change is that bjoern does not break a "keep-alive" connection without closing the peer socket. I will check this with our environment and get back with the results.

jonashaag:
No, you're confusing "number of threads/workers" and "synchronous/asynchronous I/O". Here's a good explanation: https://en.wikipedia.org/wiki/Asynchronous_I/O

jonashaag:
(Btw, envoy also uses asynchronous I/O model, but with libevent rather than libev)

BlackJohnny:
Yes indeed, you are right I was confused by your statement "can't have more clients connected than forks". Sync or async, with keep-alive, you will be limited to the number of forks.

About the fix "sock.setblocking(False)" the partial tests are looking good. I have only applied the fix on our authentication microservice and haven't received any timeout since (although this service had less timeouts in general). I will apply it tomorrow for the rest of the services and get back with feedback.

Thank you again for your time!

BlackJohnny:
PS: any planned dates for the next release? Regards!

jonashaag:
> Sync or async, with keep-alive, you will be limited to the number of forks.

No, why should that be the case? A single process can have (almost) unlimited number of connections. It can compute the response for only one connection at a time but it can be connected to many clients.

BlackJohnny:
> > Sync or async, with keep-alive, you will be limited to the number of forks.
> 
> No, why should that be the case? A single process can have (almost) unlimited number of connections. It can compute the response for only one connection at a time but it can be connected to many clients.

Yes that was clear already. My concern was about the parallel processing behavior to keep the response times acceptable. And that is dependent on how the connections are allocated across forks based on availability.  But that is my issue to solve. Thanks!

BlackJohnny:
I am getting back with the feedback on your fix after applying it to all our microservices:

**There is no timeout anymore**. :thumbsup:

Best Regards!

jonashaag:
👍

PR #171, opened 2019-11-03
filewrapper close stream before deallocate

mamico:
Call, when available, close method at the end of streaming before deallocate file-like object.

In trueness, PEP-3333 ask that wrapper implements a close() method, but IMHO close during 
`FileWrapper_Done` could be, in bjoern, equivalent.

https://www.python.org/dev/peps/pep-3333/#optional-platform-specific-file-handling
> .. the iterable returned by wsgi.file_wrapper must have a close() method that invokes the original file-like object's close() method


jonashaag:
I implemented this properly, can you try master?

mamico:
@jonashaag ok for me. thanks

PR #173, opened 2019-12-04
Fix segmentation fault in old interpreters

GoldsteinE:
Due to bug with constructing empty strings via PyString_FromFormat(),
bjoern crashed Python interpreter. It happens on Python 2 interpreters
before 2.7.16 and Python 3 interpreters before 3.5.
https://bugs.python.org/issue33817


jonashaag:
Thanks for the patch!

Can you add a comment to the two macros with a link to the bpo page.

GoldsteinE:
Done!

PR #178, opened 2020-03-31
Small step toward wheel

bersace:
Hi @jonashaag ,

Here is a first step to build wheels as requested by #121. This PR is a small step : it does not change release process, does not change tests, does not add CI, does not support every binary platform.

To upload wheel, you need docker-compose and twine. Then, simply run `make upload`. There is two new Make targets : distclean (just cleaning dist dir) and `packages` to build without uploading.

``` console
$ make distclean packages
...
$ tree dist/
dist/
├── bjoern-3.1.0-cp27-cp27mu-manylinux1_x86_64.whl
├── bjoern-3.1.0-cp27-cp27mu-manylinux2010_x86_64.whl
├── bjoern-3.1.0-cp35-cp35m-manylinux1_x86_64.whl
├── bjoern-3.1.0-cp35-cp35m-manylinux2010_x86_64.whl
├── bjoern-3.1.0-cp35-cp35m-manylinux2014_x86_64.whl
├── bjoern-3.1.0-cp36-cp36m-manylinux1_x86_64.whl
├── bjoern-3.1.0-cp36-cp36m-manylinux2010_x86_64.whl
├── bjoern-3.1.0-cp36-cp36m-manylinux2014_x86_64.whl
├── bjoern-3.1.0-cp37-cp37m-manylinux1_x86_64.whl
├── bjoern-3.1.0-cp37-cp37m-manylinux2010_x86_64.whl
├── bjoern-3.1.0-cp37-cp37m-manylinux2014_x86_64.whl
├── bjoern-3.1.0-cp38-cp38-manylinux1_x86_64.whl
├── bjoern-3.1.0-cp38-cp38-manylinux2010_x86_64.whl
├── bjoern-3.1.0-cp38-cp38-manylinux2014_x86_64.whl
└── bjoern-3.1.0.tar.gz

0 directories, 15 files
$
```

libev **IS** embedded in wheel. I'm wondering if this will trigger issue like with psycopg2 and libpq. The workaround is to use source tarball rather than wheel.

Each wheel is pen tested in a virtualenv with `test_wsgi_compliance.py`.

I hit a bug with cp27m. Instead of fighting this, I choose to blacklist it for now. Users will still use sources.

The docker-compose.yml should be easy to translate to a CI job matrix, using either CircleCI, GitHub Action, Travis, Azure pipelines and/or appveyor.

@jonashaag what do you think of this ?

cc @hoefling

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -113,8 +113,12 @@ memwatch:
 	   echo; echo; \
 	   tail -n +25 /proc/$$(pgrep -n ${PYTHON})/smaps'
 
-upload:
-	${PYTHON} setup.py sdist upload
+packages:
+	${PYTHON} setup.py sdist
+	$(MAKE) -C packaging all
+
+upload: packages
+	twine upload dist/*.tar.gz dist/*.whl
 
 $(HTTP_PARSER_OBJ):
 	$(MAKE) -C $(HTTP_PARSER_DIR) http_parser.o CFLAGS_DEBUG_EXTRA=-fPIC CFLAGS_FAST_EXTRA=-fPIC
--- a/Makefile
+++ b/Makefile
@@ -82,6 +82,9 @@ prepare-build:
 clean:
 	@rm -rf $(BUILD_DIR)/*
 
+distclean:
+	@rm -rf dist/
+
 AB		= ab -c 100 -n 10000
 TEST_URL	= "http://127.0.0.1:8080/a/b/c?k=v&k2=v2"
 
--- a/packaging/.gitignore
+++ b/packaging/.gitignore
@@ -1,0 +1,1 @@
+pip-cache/
--- a/packaging/Makefile
+++ b/packaging/Makefile
@@ -1,0 +1,4 @@
+all: manylinux2014 manylinux2010 manylinux1
+
+manylinux%:
+	IMAGE=quay.io/pypa/$@_x86_64 docker-compose run --rm builder
--- a/packaging/build_all.sh
+++ b/packaging/build_all.sh
@@ -14,8 +14,10 @@ distdir="${top_srcdir}/dist"
 teardown() {
 	# Get outer user ID and GID from self.
 	uid_gid="$(stat -c %u:%g "$0")"
-	# Ensure dist/ files are owned by user.
-	chown -R "$uid_gid" "$distdir"
+	if [ "${uid_gid}" != "0:0" ] ; then
+		# Ensure dist/ files are owned by user.
+		chown -R "$uid_gid" "$distdir"
+	fi
 }
 trap teardown INT EXIT TERM
 
--- a/packaging/build_all.sh
+++ b/packaging/build_all.sh
@@ -1,0 +1,58 @@
+#!/bin/bash -eux
+#
+# Build wheels for all python interpreter installed in /opt/python as in
+# manylinux containers.
+#
+# By default, search for a sdist tarball for version described in setup.py.
+# Accepts a specific tarball as first argument.
+#
+
+top_srcdir="$(readlink -m "$0/../..")"
+distdir="${top_srcdir}/dist"
+
+# Ensure to always clean up, even on error.
+teardown() {
+	# Get outer user ID and GID from self.
+	uid_gid="$(stat -c %u:%g "$0")"
+	# Ensure dist/ files are owned by user.
+	chown -R "$uid_gid" "$distdir"
+}
+trap teardown INT EXIT TERM
+
+cd "$top_srcdir"
+
+export PIP_DISABLE_PIP_VERSION_CHECK=off
+export PIP_NO_PYTHON_VERSION_WARNING=on
+
+runtimes=(/opt/python/cp*)
+
+version="$("${runtimes[0]}/bin/python" setup.py --version)"
+tarname="bjoern-$version"
+sdist="${1-$distdir/$tarname.tar.gz}"
+
+if ! [ -f "$sdist" ] ; then
+	echo "Missing source tarball $sdist." >&2
+	exit 1
+fi
+
+yum install -q -y libev-devel
+if ! rpm --query --queryformat= libev-devel  ; then
+	# Looks like libev-devel is not available, like in manylinux1. Let's
+	# install from sources.
+	LIBEV_VERSION=4.33
+	curl http://dist.schmorp.de/libev/Attic/libev-${LIBEV_VERSION}.tar.gz --silent --output /tmp/libev-${LIBEV_VERSION}.tar.gz
+	tar xf /tmp/libev-${LIBEV_VERSION}.tar.gz -C /tmp/
+	(cd /tmp/libev-${LIBEV_VERSION}; ./configure)
+	make -C /tmp/libev-${LIBEV_VERSION}/
+	make -C /tmp/libev-${LIBEV_VERSION}/ install
+fi
+
+# Blacklist cp27-cp27m as it triggers: site-packages/_bjoern.so: undefined
+# symbol: PyUnicodeUCS2_AsLatin1String
+runtime_blacklist=(cp27-cp27m)
+for runtimedir in "${runtimes[@]}" ; do
+	if [[ " ${runtime_blacklist[*]]} " =~ " ${runtimedir##*/} " ]] ; then
+		continue
+	fi
+	PATH="$runtimedir/bin:$PATH" $top_srcdir/packaging/build_wheel.sh "$sdist"
+done
--- a/packaging/build_wheel.sh
+++ b/packaging/build_wheel.sh
@@ -1,0 +1,43 @@
+#!/bin/bash -eux
+#
+# This script accepts a source tarball as argument and product final wheel for
+# current python interpreter. The script saves the wheel in source tarball
+# directory and pen test it to validate build.
+#
+
+sdist="$1"
+basename=${sdist/%.tar.gz}
+wheeldir=${sdist%/*}
+
+srcdir=$(readlink -m "$0/..")
+wheeltag="$(python "$srcdir/wheeltag.py")"
+impl="${wheeltag% *}"
+default_platform="${wheeltag#* }"
+
+# Build wheel from source tarball
+pip --verbose wheel --no-deps --wheel-dir "$wheeldir" "$sdist"
+wheel="$basename-$impl-$default_platform.whl"
+
+# Fix wheel to embed libev shared object.
+auditwheel repair --wheel-dir "$wheeldir" "$wheel"
+rm -f "$wheel"
+wheel="$basename-$impl-$AUDITWHEEL_PLAT.whl"
+
+# Build a venv for pentesting
+venv=/tmp/bjoern-pentestenv
+rm -rf "$venv"
+if python -m venv --help &>/dev/null ; then
+  python -m venv "$venv"
+else
+  # For Python 2.7
+  pip install virtualenv
+  hash -r
+  virtualenv "$venv"
+fi
+"$venv/bin/pip" install --upgrade pip
+
+# Pentest wheel.
+$venv/bin/pip install "$wheel"
+cd /  # Move outside source tree.
+$venv/bin/python "$srcdir/../tests/test_wsgi_compliance.py"
+rm -rf $venv
--- a/packaging/docker-compose.yml
+++ b/packaging/docker-compose.yml
@@ -1,0 +1,10 @@
+version: "3"
+
+services:
+  builder:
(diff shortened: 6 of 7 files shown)


packaging/build_wheels.sh
```diff
@@ -0,0 +1,103 @@
...
+	# Activate runtime in a subshell
+	PATH="$runtime/bin:$PATH"
+	hash -r
+
+	# Build wheel from source tarball
+	pip --verbose wheel --no-deps --wheel-dir "$distdir" "$sdist" 2>&1 | tee /tmp/wheel.log
+	wheelname="$(grep --max-count 1 --perl-regexp --only-matching "${tarname}.+\\.whl" /tmp/wheel.log)"
+
+	# Fix wheel to embed libev shared object.
+	auditwheel repair --wheel-dir "$distdir" "$distdir/$wheelname" 2>&1 | tee /tmp/wheel.log
```
jonashaag, line 20: Should probably append to `wheel.log` using `tee -a`?

jonashaag:
Looks really good. Generally speaking I dislike these kinds of sophisticated bash scripts (error handling, searching for strings in files, trap handlers, etc.) and I would suggest to write these kinds of programs in a proper programming language instead (e.g., Python). But that is no reason to rewrite the script now; it's fine.

Only concern I have is about embedded libev. Is this common practice? What are the advantages and disadvantages? I will have to give this some thought.

packaging/build_wheels.sh
bersace: Reusing the same file name may be misleading. The wheel is renamed. Thus, I need to get the filename twice for previous command output. I can update this to reduce log parsing.

bersace:
> Only concern I have is about embedded libev. Is this common practice? What are the advantages and disadvantages? I will have to give this some thought.

For the sake of portability, a wheel must not depends on something else that manylinux* defined libraries. The official `auditwheel` tools is here to embed extra libs. Here is what's in pyscopg2-binary wheel :

``` console
$ unzip -l psycopg2_binary-2.8.4-cp37-cp37m-manylinux1_x86_64.whl  | grep .libs
     9000  2019-10-20 00:57   psycopg2/.libs/libkeyutils-1-ff31573b.2.so
    61008  2019-10-20 00:57   psycopg2/.libs/liblber-2-e85f761a.4.so.2.10.11
   251816  2019-10-20 00:57   psycopg2/.libs/libsepol-b4f5b513.so.1
  3217584  2019-10-20 00:57   psycopg2/.libs/libcrypto-3a9cf061.so.1.1.1d
   167808  2019-10-20 00:57   psycopg2/.libs/libk5crypto-622ef25b.so.3.1
   738056  2019-10-20 00:57   psycopg2/.libs/libkrb5-fb0d2caa.so.3.3
     8840  2019-10-20 00:57   psycopg2/.libs/libcom_err-beb60336.so.2.1
   106904  2019-10-20 00:57   psycopg2/.libs/libselinux-cf8f9094.so.1
   341408  2019-10-20 00:57   psycopg2/.libs/libpq-3d107b13.so.5.11
   662464  2019-10-20 00:57   psycopg2/.libs/libssl-8b4167bf.so.1.1.1d
   142720  2019-10-20 00:57   psycopg2/.libs/libsasl2-6f25e95f.so.3.0.0
    87848  2019-10-20 00:57   psycopg2/.libs/libz-a147dcb0.so.1.2.3
   240936  2019-10-20 00:57   psycopg2/.libs/libgssapi_krb5-174f8956.so.2.2
    46960  2019-10-20 00:57   psycopg2/.libs/libkrb5support-d7ce89d4.so.0.1
   428744  2019-10-20 00:57   psycopg2/.libs/libldap_r-2-b688a282.4.so.2.10.11
$
```

This way, bjoern does not require users to install libev through their own package manager or by hand. Everything is pipable.

The disadvandage is where a program mixes shared objects from bjoern wheel and other wheel or host system. This is the case pyscopg2 hit with libssl. I think that libev has not the same constraint than libssl.

See https://www.psycopg.org/articles/2018/02/08/psycopg-274-released/ for details on what issues psycopg2  hit by embedding shared objects in its wheel.

bersace:
> Generally speaking I dislike these kinds of sophisticated bash scripts (error handling, searching for strings in files, trap handlers, etc.) and I would suggest to write these kinds of programs in a proper programming language instead (e.g., Python). But that is no reason to rewrite the script now; it's fine.

I fully understand this. I'm used to advanced bash scripting and find it easier when it's all about running pip, yum, make, virtualenv, setup.py, auditwheel and python interpreters.

psycopg2 did [implement this in Python](https://github.com/psycopg/psycopg2/blob/master/scripts/appveyor.py). But the difference is that pyscopg2's script is meant to run on Windows. Hence the choice was between PowerShell and Python. PowerShell and bash are very differents.

Latest fixup removes grepping the log. I wish this make you more confident.

teardown & trap is a way to ensure that script running as root in docker does not pollute git clone with root files. This is because there is (usualy) no squash of uid with docker mounted volume.

If you prefer, I could extract a `build_wheel.sh` to build one single wheel for a given runtime and sdist (mostly build_wheel function) and rework `build_wheels.sh` as `build_all.sh` calling the first script, making it simpler.

packaging/build_wheels.sh
```diff
@@ -0,0 +1,104 @@
...
+
+	# Activate runtime in a subshell
+	PATH="$runtimedir/bin:$PATH"
+	hash -r
+
+	local runtime="${runtimedir##*/}"
+
+	# Build wheel from source tarball
+	pip --verbose wheel --no-deps --wheel-dir "$distdir" "$sdist"
+	wheel="$distdir/$tarname-$runtime-linux_$AUDITWHEEL_ARCH.whl"
```
hoefling, line 19: You can query the current runtime with `setuptools.pep425tags` if you want:
```
>>> from setuptools import pep425tags
>>> pep425tags.get_impl_tag()
'cp37'
>>> pep425tags.get_abi_tag()
'cp37m'
>>> pep425tags.get_platform()
'linux_x86_64'
```

packaging/build_wheels.sh
```diff
@@ -0,0 +1,104 @@
...
+
+	# Fix wheel to embed libev shared object.
+	auditwheel repair --wheel-dir "$distdir" "$wheel"
+	rm -f "$wheel"
+	wheel="$distdir/$tarname-$runtime-$AUDITWHEEL_PLAT.whl"
+
+	# Pentest wheel in a virtualenv.
+	mkvenv /tmp/bjoern-pentestenv
+	/tmp/bjoern-pentestenv/bin/pip install "$wheel"
+	/tmp/bjoern-pentestenv/bin/python tests/test_wsgi_compliance.py
```
hoefling, line 29: Double-check whether you are really testing against installed `bjoern` here. The project doesn't use the `src` layout and you're not changing from the project root dir, so I suspect you will be importing and running code from the repository.

hoefling:
Including libs in a self-contained wheel is what the `manylinux` spec is for, so including `libev` in the fat wheel is fine IMO. As long as the source dist is offered (and one can skip fat wheels via `pip install bjoern --no-binary=bjoern`), I wouldn't expect any issues with that.

packaging/build_wheels.sh
bersace: Sadly, setuptools dropped pep425tags module. It's down deep in its _vendor directory. However, I found that `distutils.util` has a `get_platform`. I updated to remove the hardcoded `linux`. Thanks for the point !

packaging/build_wheels.sh
bersace: Good catch ! I fixed this by moving out from source tree.

packaging/build_wheels.sh
hoefling: Yeah, you're right; it's actually quite hilarious how every tool tries hiding this info (first `pip.pep425tags` moved to `pip._internal`, now `setuptools` moves it to vendored stuff). There is also `packaging.tags` available, but it has to be installed separately; the one vendored with `setuptools` is stripped AFAIK.

bersace:
@jonashaag ready for next review :-)

jonashaag:
> If you prefer, I could extract a `build_wheel.sh` to build one single wheel for a given runtime and sdist (mostly build_wheel function) and rework `build_wheels.sh` as `build_all.sh` calling the first script, making it simpler.

That would be great indeed.

Ok, it seems to be best practice to include requirements in the build, so let’s go with that. I’d be interested to know exactly why it is recommended to do it that way since it has many drawbacks (size of distribution, security, etc). I’ll try to find some sources for that info; it’s unrelated to this PR.

hoefling:
@jonashaag [PEP 513](https://www.python.org/dev/peps/pep-0513/) describes the rationale.

bersace:
> That would be great indeed.

Done.

@hoefling finally, i found that `wheel.pep425tags` is available. I wish that won't change or `wheeltag.py` script will get harder.

bersace:
@jonashaag are you working on a Mac ? Does it work properly on your environment ?

jonashaag:
Yes, I have multiple Mac machines, will try soon.

bersace:
@jonashaag If you want to upload wheels for older release, I can tweak it a bit to have the following post-release packaging process:

- Clean with `make cleandist`
- Download previous sdist from PyPI in dist/
- `make -C packaging all`
- `twine upload dist/*.whl`

jonashaag:
Somehow, I totally forgot about this. Sorry. Testing it now.

jonashaag:
Build seems to work, but got a problem with the `chmod` in `teardown`. Maybe that's Mac specific.

```
++ stat -c %u:%g /workspace/packaging/build_all.sh
+ uid_gid=0:0
+ chown -R 0:0 /workspace/dist
make[1]: *** [manylinux2014] Error 1
make: *** [packages] Error 2
```

UID/GID will always be 0.

bersace:
> Build seems to work, but got a problem with the chmod in teardown. Maybe that's Mac specific.

Right. When bind mounting volumes, it's better to sync ownership with git checkout so that upload tools have access and it's easy to cleanup dist directory.

I'll add a check on uid to avoid chmoding 0:0. Do you have a better trick ?

bersace:
@jonashaag fixup pushed. Would you mind to test it ? Please wait the rebase before merging :-)

jonashaag:
It builds ok now! Will review more asap

bersace:
:up:

jonashaag:
Sorry for the huge latency. I was mostly busy with other things, and also not being familiar with wheel packaging and advanced bash scripting didn't help. But finally got around to reviewing this and from what I understand it's great! Thanks!

jonashaag:
When I run on Linux, this is what I get:

```
IMAGE=quay.io/pypa/manylinux2014_x86_64 docker-compose run --rm builder
Creating packaging_builder_run ... done
++ readlink -m /workspace/packaging/build_all.sh/../..
+ top_srcdir=/workspace
+ distdir=/workspace/dist
+ trap teardown INT EXIT TERM
+ cd /workspace
+ export PIP_DISABLE_PIP_VERSION_CHECK=off
+ PIP_DISABLE_PIP_VERSION_CHECK=off
+ export PIP_NO_PYTHON_VERSION_WARNING=on
+ PIP_NO_PYTHON_VERSION_WARNING=on
+ runtimes=(/opt/python/cp*)
++ /opt/python/cp310-cp310/bin/python setup.py --version
+ version=3.1.0
+ tarname=bjoern-3.1.0
+ sdist=/workspace/dist/bjoern-3.1.0.tar.gz
+ '[' -f /workspace/dist/bjoern-3.1.0.tar.gz ']'
+ echo 'Missing source tarball /workspace/dist/bjoern-3.1.0.tar.gz.'
Missing source tarball /workspace/dist/bjoern-3.1.0.tar.gz.
+ exit 1
+ teardown
++ stat -c %u:%g /workspace/packaging/build_all.sh
+ uid_gid=1000:0
+ '[' 1000:0 '!=' 0:0 ']'
+ chown -R 1000:0 /workspace/dist
chown: cannot access ‘/workspace/dist’: No such file or directory
ERROR: 1
make: *** [Makefile:4: manylinux2014] Error 1
```

bersace:
Hi @jonashaag the scripts requires tarball :

```
...
+ echo 'Missing source tarball /workspace/dist/bjoern-3.1.0.tar.gz.'
Missing source tarball /workspace/dist/bjoern-3.1.0.tar.gz.
...
```

I can submit a patch to add better logging rather than bash verbose '+'. Also, the final chown fails if no dist dir, but that's not the root of the error.

jonashaag:
Ah, didn't see that line!

PR #195, opened 2020-11-13
Update gitmodules to use HTTPS URL for http-parser

parazyd:
This commit changes the submodule URL to a secure one.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "http-parser"]
   path = http-parser
-  url = git://github.com/joyent/http-parser
+  url = https://github.com/joyent/http-parser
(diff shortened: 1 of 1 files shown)


jonashaag:
Thanks! What's the advantage of that?

parazyd:
Avoiding MITM when cloning, and also some package managers prefer it, and give noisy warnings otherwise.
There's no downside to this change really.

jonashaag:
I’m not sure if it will break people’s checkouts. Do you know?

parazyd:
It won't break anything. New clones will simply use the https URI, and old checkouts which use the git URI are already configured in `.git/modules`, so nothing will change there.

parazyd:
Thanks!

Issue #196, opened 2020-11-23
Broken pipe error(errno 32) is looping for unix sockets


supratikchatterjee16:
While testing an application there seems to be an issue that keeps arising and fills my error logs by
adding the following errors : 

![image](https://user-images.githubusercontent.com/25512106/99941962-e6788580-2d94-11eb-9b1d-a3acc07e944a.png)

My code : 
```python
def deploy():
	import os
	import app
	import bjoern
	import signal
	import socket
	host = '0.0.0.0'
	# port = 8766 # W(87)B(66) for Workbench
	NUM_WORKERS = 10
	worker_pids = []
	socket_path = './something.sock'
	sock = socket.socket(socket.AF_UNIX)
	sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
	sock.settimeout(1) # Assumption seconds
	sock.bind(socket_path)
	sock.listen(1024)
	os.chmod(socket_path, 0o666)
	try:
		print('Attempting run : ')
		bjoern.server_run(sock, app.create())
	except KeyboardInterrupt:
		print('Eliminating Unix Socket')
		os.unlink(sock.getsockname())
		sock.close()
	except Exception as e:
		print(e)
```

I recreate the error by running the program and then continuously refreshing the page, 
but it starts doing this automatically as well.

The line being executed(permalink):

https://github.com/jonashaag/bjoern/blob/f50e514885c10793ec7cfce845015ac17e82ef69/bjoern/server.c#L521

It isn't hanging up as you intended it to be, I think.

jonashaag:
Does this also happen if you use he built in socket initialization? What about port based sockets?

supratikchatterjee16:
It does not happen with the TCP port based sockets.
But is happening even with the inbuilt Unix sockets intialization.

```python
# Tested lines
bjoern.run(app.create(), 'localhost', 8000, reuse_port = True)
bjoern.run(app.create(), 'unix:'+os.path.join(os.getcwd(), 'workbench.sock'))
```

jonashaag:
Can't reproduce, can you try with this app? How exactly are you sending the request? Can you try with cURL?

```
def app(environ, start_response):
    start_response('200 OK', [])
    return ''
```

```
curl -v --unix-socket ./something.sock /someurl
```

supratikchatterjee16:
I modified the test scripts a bit to match the situation, it is recurring, albeit slowly this time :

The script : 

```shell
[...]

PR #197, opened 2020-11-29
FIx for looping EPIPE issue for unix sockets

supratikchatterjee16:
@jonashaag 

Made modifications to prevent Issue #196 .

Tested for inbuilt TCP stream and unix sockets on Linux only.
Do review the code whenever possible.


ayalster:
@jonashaag @supratikchatterjee16 
Hi, could you please share if there is a particular reason why this PR was closed and not merged to master?
As far as i can tell the issue #196  still exists and nothing in the current realese address it. Am i missing something?
I do think i encouter it when using version 3.2.2.

jonashaag:
I don’t remember. Does it fix things for you?

ayalster:
Sorry for the late reply - 
It solves a standing issue that i encounter where, when connected via unix sockets, if the client disconnects before reply - it triggers an infinite error loop. I think it may be fixed in a different way in the fixes branch, but i don't know when if / when it is supposed to be merged into a new release version.

jonashaag:
Did you actually try the branch?

ayalster:
Do you mean the fixes branch? no - i preferred to stick with the tried and tested release. and it seemed to have too many changes from master.

jonashaag:
Would be good to know if the fix in this PR actually fixes the problem!

ayalster:
Yes, indeed it solves the issue

supratikchatterjee16:
@jonashaag due to the repository having grown stale by more than a year, I had closed it in order to clean up my own repository. The fix was to implement an unimplemented mechanism that had been coded in.

Issue #198, opened 2020-12-06
A Docker image


x-yuri:
In #153 you suggested to create a Docker image. Supposedly to be able to run `bjoern`. Here's the starting point:

```dockerfile
FROM python:alpine
RUN apk add --no-cache build-base libev-dev git \
    && cd /usr/src \
    && git clone https://github.com/jonashaag/bjoern \
    && cd bjoern \
    && git submodule update --init \
    && python setup.py install
```

```
$ docker build -t bjoern .
```

I could download the latest release:

```
    && wget https://github.com/jonashaag/bjoern/archive/$VERSION.tar.gz -O bjoern.tar.gz \
    && tar xf bjoern.tar.gz -C /usr/src \
    && cd /usr/src/bjoern-*
```

But it doesn't come with submodules. Although it contains `.gitignore`, and `.gitmodules`, so if I do `git init`... that would probably still fail.

And what do you expect to happen when one does `docker run --rm bjoern`? And how do I feed an application to it? With e.g. `nginx`, by default it shows the welcome page. And to put it to practical use you replace the config, and add a directory with static files.

jonashaag:
Cool! The default Docker image doesn't have to cover advanced use cases. I'd say we should expected the application to be mounted to `/app`. I just put together a simple runner script (maybe we can add it too the main repo as well...) that we can add to the Docker image and use as entry point:

```py
import argparse
import functools
import importlib
import multiprocessing

import bjoern


def load_app(app_module):
    try:
        app_module_name, app_obj_name = app_module.rsplit(":", 1)
    except ValueError:
        app_module_name = app_module
        app_obj_name = "application"
    app_module = importlib.import_module(app_module_name)
    try:
        app_obj = getattr(app_module, app_obj_name)
    except AttributeError as err:
        raise ImportError(
            f"cannot import name {app_obj_name!r} from {app_module_name!r} ({app_module.__file__})"
        ) from err
    return app_obj


def main():
    parser = argparse.ArgumentParser()
    parser.add_argument("--bind", "-b", type=str, default="127.0.0.1:8000")
    parser.add_argument("--workers", "-w", type=int, default=1)
    parser.add_argument("app_module")
    args = parser.parse_args()

    app_obj = load_app(args.app_module)

    host, port = args.bind.rsplit(":")
    port = int(port)

    workers = [
        multiprocessing.Process(
            target=functools.partial(bjoern.run, app_obj, host, port, reuse_port=True),
            daemon=True,
        )
[...]

PR #204, opened 2020-12-31
Run an application via the cli

Julien00859:
Some people don't want to embed bjoern directly in their WSGI app
but instead let it run on any WSGI server. Developers usually prefer not
to run a fullblow server (which can be resource consuming, slow to
start/restart) but instead run a simple, embeded web server like the
python standard wsgiref.

Following the WSGI specification, if one application implement client
side of the protocol then any server that implement the server side
should be able to run it.

Prior to this commit, the prefered way to use bjoern as web server was
to embed it directly in the source code of the application, this defeat
the two above situation, it forces developers to build bjoern on their
machine even if they don't need the perfs of bjoern, it also prevent any
other WSGI server to be used. In other words, bjoern is a library the
developers should embed in their application, it is the application that
loads and run bjoern.

Thanks to this commit, bjoern can be used as a standalone process that
load the WSGI application as part of its bootup. The CLI consumes only a
few options that are used by bjoern (IP, port, unix domain socket, reuse
address, statsd), the other unknown options are left to be parsed by the
application.

Using bjoern as a standalone process, it is no more required to load it
as part of the application. The application becomes WSGI-server
independant.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -34,6 +34,8 @@ Flask example
 
 .. code-block:: python
 
+   # flask_example.py
+
    from flask import Flask
 
    app = Flask(__name__)
@@ -46,6 +48,13 @@ Flask example
        import bjoern
 
        bjoern.run(app, "127.0.0.1", 8000)
+
+
+Alternatively, start the application via the CLI:
+
+.. code-block:: bash
+
+   $ python -m bjoern --host 127.0.0.1 --port 8000 flask_example:app
 
 
 Advanced usage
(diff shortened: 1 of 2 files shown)


Julien00859:
Branch force-pushed to add py2 and py34 compat

The diff:

    -    sys.argv = [sys.argv[0], *other_args]
    +    sys.argv = [sys.argv[0]]
    +    sys.argv.extend(other_args)

Julien00859:
I did set the +x permission on the file and added the shebang to python3

jonashaag:
See also https://github.com/jonashaag/bjoern/issues/198#issuecomment-751378298

Julien00859:
I love the `load_app` you wrote in \#198, definitely better than my two-liners

PR #219, opened 2021-08-13
Fix #218: SERVER_PORT correctly set

### LINKED ISSUE #218: SERVER_PORT and SERVER_HOST values are incorrect
With this code

```
if __name__ == "__main__":
    import bjoern

    bjoern.run(app, host="localhost", port=8000)
```

I get : 
'SERVER_NAME': '127.0.0.1',
'SERVER_PORT': '1596915664'

Each time I run my server, whatever the initial values are, I get a random int as a port. What's going on ?

trollfot:
This probably needs at least a test but I couldn't figure out how the tests are ran.


jonashaag:
Thanks, we don't really have unit tests so this is fine.

PR #236, opened 2022-08-28
Implement graceful shutdown

### LINKED ISSUE #91: Gracefully exiting a Python process running Bjoern

nhoad:
Hi! This PR attempts to implement per the recommendation in this comment https://github.com/jonashaag/bjoern/issues/91#issuecomment-493710529. The commit message provides more detail on exactly

Note that one distinct difference from the suggested implementation is that the listener socket is closed after the backlog is drained. For my use case (running many instances of my HTTP server using SO_REUSEPORT) this works better, as it will ensure that any new connections will go to other instances ASAP and ensure this given instance can shutdown relatively quickly. Based on further discussion on the issue, I think this would be the most beneficial behavior for others too. If you'd like that behavior to be configurable as well, just let me know and I can add another config flag for it.

(hopefully) fixes #91.


bjoern/server.c
```diff
@@ -141,46 +196,81 @@ ev_timer_ontick(struct ev_loop* mainloop, ev_timer* watcher, const int events)
...
+#ifdef WANT_GRACEFUL_SHUTDOWN
+static void
+ev_shutdown_ontick(struct ev_loop* mainloop, ev_timer* watcher, const int events)
+{
+  ThreadInfo *thread_info = (ThreadInfo*)ev_userdata(mainloop);
+  DBG("ev_shutdown_ontick %d", thread_info->server_info->active_connections);
+
+  if (thread_info->server_info->active_connections == 0) {
+    DBG("No more active connections, shutting down");
+    GIL_LOCK(0);
```
jonashaag, line 208: Why do we need the GIL here?

bjoern/server.c
```diff
@@ -141,46 +196,81 @@ ev_timer_ontick(struct ev_loop* mainloop, ev_timer* watcher, const int events)
...
-    return;
-  }
-
-  int flags = fcntl(client_fd, F_GETFL, 0);
-  if(fcntl(client_fd, F_SETFL, (flags < 0 ? 0 : flags) | O_NONBLOCK) == -1) {
-    STATSD_INCREMENT("conn.accept.error");
-    DBG("Could not set_nonblocking() client %d: errno %d", client_fd, errno);
-    return;
-  }
+  while ((client_fd = accept(watcher->fd, (struct sockaddr*)&sockaddr, &addrlen)) > 0) {
```
jonashaag, line 246: Is it safe to accept multiple requests in a single callback call? If yes, suggest to rename to `ev_io_on_requests`

bjoern/server.c
```diff
@@ -141,46 +196,81 @@ ev_timer_ontick(struct ev_loop* mainloop, ev_timer* watcher, const int events)
...
+    DBG("Accepted %d", client_fd);
+    int flags = fcntl(client_fd, F_GETFL, 0);
+    if(fcntl(client_fd, F_SETFL, (flags < 0 ? 0 : flags) | O_NONBLOCK) == -1) {
+      STATSD_INCREMENT("conn.accept.error");
+      DBG("Could not set_nonblocking() client %d: errno %d", client_fd, errno);
+      return;
+    }
 
-  GIL_LOCK(0);
+    GIL_LOCK(0);
```
jonashaag, line 255: Let's acquire the GIL outside the loop, it's quite costly.

bjoern/server.c
jonashaag, line 246: Is it safe to accept multiple requests in a single callback call? If yes, suggest to rename to `ev_io_on_requests`

bjoern/_bjoernmodule.c
```diff
@@ -11,6 +11,10 @@ static PyObject*
 run(PyObject* self, PyObject* args)
 {
   ServerInfo info;
+#ifdef WANT_GRACEFUL_SHUTDOWN
+  info.active_connections = 0;
```
jonashaag, line 15: Is there a built in way in libev to handle connection draining so that we don't have to keep a counter? I'm worried about counting bugs.

bjoern/server.c
```diff
@@ -123,14 +141,51 @@ ev_signal_on_sigint(struct ev_loop* mainloop, ev_signal* watcher, const int even
...
+  ThreadInfo *thread_info = (ThreadInfo*)ev_userdata(mainloop);
+
+  ev_io_stop(mainloop, &thread_info->accept_watcher);
   ev_signal_stop(mainloop, watcher);
 #ifdef WANT_SIGNAL_HANDLING
   ev_timer_stop(mainloop, &timeout_watcher);
 #endif
+#ifdef WANT_GRACEFUL_SHUTDOWN
+  ev_signal_stop(mainloop, &thread_info->sigterm_watcher);
+  ev_timer_stop(mainloop, &timeout_watcher);
```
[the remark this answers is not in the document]
nhoad: Good point! I'll move the timeout watcher up to the top and shuffle the code around so that it only gets stopped once.

From the documentation for `ev_TYPE_stop` here http://pod.tst.eu/http://cvs.schmorp.de/libev/ev.pod, it says "Stops the given watcher if active, and clears the pending status (whether the watcher was active or not)." so it looks like it's safe to stop it more than once. Because it also clears the pending status means that we shouldn't receive a `timeout_watcher` callback during this handler either.

bjoern/server.c
```diff
@@ -123,14 +141,51 @@ ev_signal_on_sigint(struct ev_loop* mainloop, ev_signal* watcher, const int even
...
+#endif
 }
 #endif
 
+#ifdef WANT_GRACEFUL_SHUTDOWN
+static void
+ev_signal_on_sigterm(struct ev_loop* mainloop, ev_signal* watcher, const int events)
+{
+  ThreadInfo *thread_info = (ThreadInfo*)ev_userdata(mainloop);
```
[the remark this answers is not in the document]
nhoad: Good point! Will do.

bjoern/server.c
nhoad: Oh yeah! We definitely don't. I'll remove this.

bjoern/server.c
nhoad: Yep! At this point it's just regular UNIX socket stuff, so accepting as much as you can is akin to reading as much as you can. I'll rename the function.

bjoern/_bjoernmodule.c
nhoad: Unfortunately no, going through the documentation I can't see any way to do that. In case it helps, I went through a couple of iterations of the counting code and this was by far the simplest one I came up with - counting connections as "active" once http-parser tells us a message has begun, and then not counting them once we're done with the sending the response.

I contemplated doing something like... storing all of the requests in an array/linked list/hash map on the ThreadInfo struct, and then basically watching for that to drop to either drop to zero entries, or if the existing entries aren't handling any connections. That would eliminate counting bugs, but it would also be significantly more complex as we'd need to handle growing the array occasionally if there's an influx of requests. I think we'd also need to store some additional state on the Request struct for whether or not it's currently handling a request. Given the complexity, that's why I decided to go for the simpler approach :)

That said though, I think it's worth doing some testing on dropping connections halfway through requests at various points in the lifecycle to see how the counting holds up. I'll go back and do that, and report on my findings :)

nhoad: My testing found something - aborted connections weren't being accounted for. I've fixed that up and added some tests for it as well.

bjoern/server.c
```diff
@@ -141,46 +196,76 @@ ev_timer_ontick(struct ev_loop* mainloop, ev_timer* watcher, const int events)
...
-    client_fd,
-    inet_ntoa(sockaddr.sin_addr)
-  );
+  while ((client_fd = accept(watcher->fd, (struct sockaddr*)&sockaddr, &addrlen)) > 0) {
+    DBG("Accepted %d", client_fd);
+    int flags = fcntl(client_fd, F_GETFL, 0);
+    if(fcntl(client_fd, F_SETFL, (flags < 0 ? 0 : flags) | O_NONBLOCK) == -1) {
+      STATSD_INCREMENT("conn.accept.error");
+      DBG("Could not set_nonblocking() client %d: errno %d", client_fd, errno);
+      return;
```
jonashaag, line 249: Missing GIL unlock. Might want to use `goto out` or something like that.

bjoern/_bjoernmodule.c
jonashaag: Why not implement counting as follows:

- A successful `accept` increments the counter
- A `close` decrements the counter

So, move to `ev_io_on_requests` and `close_connection`?

bjoern/server.c
```diff
@@ -117,20 +138,54 @@ pyerr_set_interrupt(struct ev_loop* mainloop, struct ev_cleanup* watcher, const
...
+{
+
+  ev_io_stop(mainloop, &THREAD_INFO->accept_watcher);
+  // Drain the accept queue before we close to try and minimize connection resets
+  ev_io_on_requests(mainloop, &THREAD_INFO->accept_watcher, 0);
+  // Close the socket now to ensure no more clients can talk to us
+  close(THREAD_INFO->server_info->sockfd);
+
+  ev_signal_stop(mainloop, watcher);
+  ev_timer_stop(mainloop, &timeout_watcher);
```
jonashaag, line 171: Move this to the top as well?

bjoern/server.c
```diff
@@ -117,20 +138,54 @@ pyerr_set_interrupt(struct ev_loop* mainloop, struct ev_cleanup* watcher, const
...
+
+  // don't shutdown immediately, but start a timer to check if we can shutdown in 100ms.
```
jonashaag, line 173: Do we still need this if we change the counting to what I suggested above? It feels like this timeout stuff is unnecessarily complicated

bjoern/server.c
nhoad: Oof, that would have been bad! Great catch.

bjoern/_bjoernmodule.c
nhoad: We could do that! That would certainly be simpler. The downside to doing that is that any persistent connections that aren't being actively used will hold the event loop open until the timeout is reached. Is that acceptable?

bjoern/server.c
nhoad: We would still need it, yes - without this there's nothing checking if there are any active connections, so it will always take 30 seconds to shutdown (i.e. the other timeout would be reached).

nhoad: Thinking a little bit more on this, I _could_ remove the `ev_shutdown_timeout_ontick` function and associated timer if I did the following:

- When SIGTERM is received, store `ev_now(loop) + 30` in an `ev_tstamp` on the server
- Make `ev_shutdown_ontick` check if `shutdown_time <= ev_now(loop)`, and use that to stop the loop.

That feels a bit simpler to me than timeout juggling it's currently doing.

jonashaag:
@nhoad are you planning to work on this some more?

nhoad:
I'm happy with the branch as it is, but if you'd like me to make any changes just let me know 😄

PR #248, opened 2024-08-09
wip single chunk merging

jonashaag:


jonashaag:
@Pebin check out this branch

Pebin:
Unfortunately I don't see any difference between this version and the original one. It happened to me randomly that maybe 1 in 10 try finished as expected (2 responses after one second) but otherwise the behavior seems to be still the same 

or it could be that I am building it/using it wrong. But I even renamed the version just in case and I see the correct version installed in the virtual env :thinking:

